PR #1, opened 2022-06-27
Implement Across V2 events scraper

amateima:
### Description

This PR represents a more centralised approach of scraping on-chain & off-chain data relevant to the Across protocol.
With the increased number of deposits it became clear that some particular features of the application can't be implemented by fetching the events and their associated data in browser at each page load. 

This is Across Scraper API, a web service that aggregates data from the chain (blocks, transactions, events) and other off chain sources (CoinGecko), in order to compute the state of the deposits happening on the protocol. 
This service will be used for:
1. All TX History Page
2. Rewards/Referral Page and other upcoming features
3. Other apps/tools that may want cached data about blocks, transactions, etc
4. TBD

This first PR contains the pipeline used to aggregate the data. This is the foundation for the API layer which will be added as a separate PR.

This implementation started by migrating the implementation of the TX History page from the FE to a dedicated web service, but with a series of improvements to increase its stability and performance. As an overview of the implementation, it consists of a few, but well defined steps:
1. Identify the blocks that have to be queried
2. Fetch the fill and deposit events corresponding the the queried block range
4. Save the deposit and fill events
5. Make additions in order to have a complete dataset (getBlock for deposit timestamp, getTransaction for referral address, get ERC20 token details, get CoinGecko price, etc)

### Tech stack

This service is implemented using (NestJS)[https://nestjs.com/], a web framework built on top of ExpressJS. In order to make it reliable and fault tolerant I chose to use PostgreSQL and Redis for managing the data. 
- **PostgreSQL** is meant to be used as a warehouse for storing the deposits data, along with relevant information about blocks, transactions, prices and other data. I'm already running this locally, meaning that the production `block` and `transaction` table can be pre-populated with the data I already have locally. This makes a lot of `getBlock` and `getTransaction` calls to be skipped
- **Redis** was added as a persistence layer for the queues that are used for processing the data. All the steps of the process were isolated as separate queues that are persistent and retryable. This means that in case of server restarts/deployments/errors, the queue messages will be retried with no need to fetch the data again from the initial source.

### Implementation

This schema describes how messages are distributed from one queue to another. All the queue messages are retried until the queue worker/consumer finishes the execution successfully. 

Protocol UMA-6

1. The first step is to identify the blocks that need to be queried. `from` depends on the latest block that was queried **OR** the block number configured to be used as starting point (to avoid querying the chain from block 0). `to` usually represents the latest block from the chain, but it can be lower to ensure that `to - from` is limited to a max number of blocks [defined here](https://github.com/across-protocol/scraper-api/pull/1/files#diff-bb8f574ca10ada36c4348f8454670dbfb452ba295bfafe7e0ca187b893590a8bR100-R106). The `{ chainId, from, to }` messages are pushed to the `BlocksEvents` queue

2. **BlocksEvents queue**
The messages submitted to this queue are processed by [BlocksEventsConsumer](https://github.com/across-protocol/scraper-api/pull/1/files#diff-61ae158016fc8c09b96a033dcc351eff583b6107c61941d1b59d768ffa563ff7R29). Its job is to fetch the deposit and fill events, save the Deposit events in the DB, add the Deposit events to the `BlockNumber` queue and add the Fill events to a [dedicated queue](https://github.com/across-protocol/scraper-api/pull/1/files#diff-2f301dbc92552cf0aa8dd8f2a4c98c01b3966f302e5cf706b2b0f6ddebaff768R21) to be inserted in the database.

🤝 If a particular block range needs to be re-submitted on this queue for processing it again, we can submit messages using the API endpoints. 

3. **BlockNumber queue** 
The messages submitted to this queue are processed by the [BlockNumberConsumer](https://github.com/across-protocol/scraper-api/pull/1/files#diff-834eca46c3f0a8e75bb2e3fbf8723790176a5b13ce3f9979baf27853e3657a23R22). During this processed we fetch the block (from the database or from the DB in case it was saved previously) that's used to determine exact the deposit time. In the end, the deposit identifier is sent to the `TokenDetails` and `DepositReferralConsumer`

4. **TokenDetails queue** and **DepositReferral queue**
The messages submitted to these queues are processed [here](https://github.com/across-protocol/scraper-api/pull/1/files#diff-1e7aa1b623afa7fea473e3682202d8c8b597986bf01a5dbc04a9a1b2e777c290R20) and [here](https://github.com/across-protocol/scraper-api/pull/1/files#diff-05917c87ba2682370a74e5b935e4d58c504d2887ee72c614eebcb24d820da9fcR21). The goal is to fetch additional data about the deposit: ERC20 token details and the optional referral address

5. **FillEvents queue**
The Fill events are inserted asynchronously in [this consumer](https://github.com/across-protocol/scraper-api/pull/1/files#diff-2f301dbc92552cf0aa8dd8f2a4c98c01b3966f302e5cf706b2b0f6ddebaff768), as the corresponding Deposit event may be missing from the database. These messages are retried until the corresponding Deposit event is found in the database.


nicholaspai (approved):
LGTM this is a great first start and I understand the architecture. Will review closer the next follow up PR's

PR #8, opened 2022-07-08
feat: add multiple referral address strategies

amateima:
Signed-off-by: amateima <amatei@umaproject.org>

This PR adds support for identifying referral addresses from calldata that contains delimiter bytes when a configured block number is passed. Until the configured block number we need to also check the nonce of the extracted referral address. 

https://linear.app/uma/issue/ACX-7/filter-out-referrer-addresses


.env.sample
```diff
@@ -11,4 +11,5 @@ WEB3_NODE_URL_1=https://infura.io
 WEB3_NODE_URL_10=https://infura.io
 WEB3_NODE_URL_288=https://infura.io
 WEB3_NODE_URL_42161=https://infura.io
-WEB3_NODE_URL_137=https://infura.io
\ No newline at end of file
+WEB3_NODE_URL_137=https://infura.io
+REFERRAL_DELIMITER_START_TIMESTAMP=9999999999999
```
amateima, line 15: pass the block number as env variable

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -15,44 +17,36 @@ export class DepositReferralConsumer {
...
+    let referralAddress: string | undefined = undefined;
+
+    if (referralDelimiterStartTimestamp && blockTimestamp >= referralDelimiterStartTimestamp) {
+      referralAddress = this.referralService.extractReferralAddressUsingDelimiter(transaction.data);
+    } else {
+      referralAddress = this.referralService.extractReferralAddress(transaction.data);
+
+      if (referralAddress) {
+        const nonce = await this.ethProvidersService.getProvider(ChainIds.mainnet).getTransactionCount(referralAddress);
+        if (nonce === 0) referralAddress = undefined;
```
amateima, line 46: also check the nonce if we found referral address prior to the configured block number

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -15,44 +17,36 @@ export class DepositReferralConsumer {
...
       }
     }
-    return undefined;
+    await this.depositRepository.update({ id: deposit.id }, { referralAddress: referralAddress || null });
```
amateima, line 49: always update row in the table in case an existing deposit row is processed

src/modules/scraper/entry-point/http/dto.ts
```diff
@@ -14,3 +14,13 @@ export class ProcessBlocksBody {
...
+
+export class SubmitReferralAddressJobBody {
+  @IsInt()
+  @ApiProperty({ example: 1 })
+  fromDepositId: number;
+
+  @IsInt()
+  @ApiProperty({ example: 2 })
+  toDepositId: number;
+}
```
amateima, line 26: validation of the POST request body

src/modules/referral/services/service.ts
```diff
@@ -71,4 +75,50 @@ export class ReferralService {
...
     return { referralRate: 0.4, tier: 1 };
   }
+
+  public subtractFunctionArgsFromCallData(data: string) {
+    const coder = new ethers.utils.AbiCoder();
+    // strip hex method identifier
+    const dataNoMethod = ethers.utils.hexDataSlice(data, 4);
+    // keep method hex identifier
+    const methodHex = data.replace(dataNoMethod.replace("0x", ""), "");
+    const decodedData = coder.decode(
```
nicholaspai, line 85: Would this work if the transaction is sent to a non Across contract, like what Socket uses? They use some other contract that calls Across's spoke pool

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
nicholaspai: +1

nicholaspai (approved):
Some small questions, lgtm

src/modules/referral/services/service.ts
amateima: I had the same question. I didn't test it extensively because I saw that all the deposits were getting filled in the database. This means that all the event parameters: deposit id, chain id, amount, etc were decoded correctly.

src/modules/referral/services/service.ts
```diff
@@ -71,4 +75,47 @@ export class ReferralService {
...
+    const fullEncoded = methodHex + encoded.replace("0x", "");
+    return data.replace(fullEncoded, "");
+  }
+
+  public extractReferralAddressUsingDelimiter(data: string) {
+    const referralData = this.subtractFunctionArgsFromCallData(data);
+
+    if (referralData.indexOf(REFERRAL_ADDRESS_DELIMITER) !== -1) {
+      const addressIndex = referralData.indexOf(REFERRAL_ADDRESS_DELIMITER) + REFERRAL_ADDRESS_DELIMITER.length;
+      const potentialAddress = referralData.slice(addressIndex, addressIndex + 40);
```
mrice32, line 99: Should we do this instead to also weed out cases where there are _more than_ 40 additional characters?

```suggestion
      const potentialAddress = referralData.slice(addressIndex);
```

src/modules/web3/services/EthProvidersService.ts
```diff
@@ -99,8 +99,13 @@ export class EthProvidersService {
 
     if (!transaction) {
       const web3Transaction = await this.getProvider(chainId).getTransaction(hash);
-      const { data } = web3Transaction;
-      transaction = this.transactionRepository.create({ chainId, data, hash });
+      const { data, blockNumber } = web3Transaction;
```
mrice32, line 102: Are we planning to use this database for other services? The only reason I ask is that the stored data is somewhat limited for each of the types (transactions, blocks, especially), and I wanted to better understand how painful the upgrade process will be once this is in production and we decide that we want to add a new transaction field to the db (for instance).

Put differently, will we be building a separate transaction, block, and event database for each service that we run (referrals, across API, transactions page)? Or will we be building a unified API that serves all of them through the same DB that is easy to upgrade when we need a new field.

Note: this may just be my ignorance about how SQL databases are typically managed.

mrice32 (approved):
Looks good! A few minor questions

src/modules/web3/services/EthProvidersService.ts
amateima: @mrice32 It is a unified API that will use the same tables. The table fields are easy to update at a later time 👍

src/modules/referral/services/service.ts
amateima: @mrice32  I took into consideration the fact that in some cases the referral address can be found at other positions than the end of the input string 🤷‍♂️

PR #10, opened 2022-07-12
fix: make ACX price 0.1$

amateima:
Signed-off-by: amateima <amatei@umaproject.org>

When computing ACX rewards, the token price should be 0.1$

https://linear.app/uma/issue/ACX-42/change-acx-token-price-to-01dollar

### DIFF
--- a/src/modules/configuration/index.ts
+++ b/src/modules/configuration/index.ts
@@ -58,5 +58,6 @@ export default registerAs("config", () => {
         },
       },
     },
+    acxUsdPrice: 0.1,
   };
 });
--- a/src/modules/referral/services/queries.ts
+++ b/src/modules/referral/services/queries.ts
@@ -4,10 +4,10 @@ export const getReferralsQuery = () => {
       *,
       case
         when d."depositorAddr" = $1 and d."referralAddress" = $1
-          then cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / 0.025 * power(10, 18) as varchar)
+          then cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / $2 * power(10, 18) as varchar)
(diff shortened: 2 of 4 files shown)


src/modules/configuration/index.ts
```diff
@@ -58,5 +58,6 @@ export default registerAs("config", () => {
         },
       },
     },
+    acxUsdPrice: 0.1,
```
amateima, line 61: add ACX price in the config so it can be changed at any time

src/modules/referral/services/queries.ts
```diff
@@ -4,10 +4,10 @@ export const getReferralsQuery = () => {
       *,
       case
         when d."depositorAddr" = $1 and d."referralAddress" = $1
-          then cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / 0.025 * power(10, 18) as varchar)
+          then cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / $2 * power(10, 18) as varchar)
         when d."depositorAddr" = $1
-        then cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / 0.025 * 0.25 * power(10, 18) as varchar)
-        else cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / 0.025 * 0.75 * power(10, 18) as varchar)
+        then cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / $2 * 0.25 * power(10, 18) as varchar)
```
amateima, line 9: The price is not hardcoded anymore in the SQL query, but received as a dynamic parameter coming from the global config

nicholaspai: Why $2?

nicholaspai (approved):
Looks good!

PR #13, opened 2022-07-18
fix: change query cast type

amateima:


nicholaspai (approved):

src/modules/referral/services/queries.ts
```diff
@@ -4,10 +4,10 @@ export const getReferralsQuery = () => {
       *,
       case
         when d."depositorAddr" = $1 and d."referralAddress" = $1
-          then cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / $2 * power(10, 18) * d.multiplier as varchar)
+          then cast(d."realizedLpFeeUsd" * (1 + d."referralRate") / $2 * power(10, 18) * d.multiplier as decimal)
```
amateima, line 7: wei numbers should be cast as decimals, not varchar

PR #14, opened 2022-07-18
fix: fix bug on fetching token prices

amateima:
There's a bug on the logic that fetches token prices. The current date was used instead of deposit date

### DIFF
--- a/src/modules/scraper/adapter/messaging/TokenPriceConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/TokenPriceConsumer.ts
@@ -22,9 +22,9 @@ export class TokenPriceConsumer {
     const { depositId } = job.data;
     const deposit = await this.depositRepository.findOne({ where: { id: depositId }, relations: ["token"] });
 
-    if (!deposit) throw new Error("Deposit not found");
-    if (!deposit.tokenId || !deposit.token) throw new Error("Deposit has not token");
-    const previousDate = DateTime.now().minus({ days: 1 }).toJSDate();
+    if (!deposit) return;
+    if (!deposit.tokenId || !deposit.token || !deposit.depositDate) throw new Error("Invalid deposit");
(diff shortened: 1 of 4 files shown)


src/modules/scraper/adapter/messaging/TokenPriceConsumer.ts
```diff
@@ -22,9 +22,9 @@ export class TokenPriceConsumer {
     const { depositId } = job.data;
     const deposit = await this.depositRepository.findOne({ where: { id: depositId }, relations: ["token"] });
 
-    if (!deposit) throw new Error("Deposit not found");
-    if (!deposit.tokenId || !deposit.token) throw new Error("Deposit has not token");
-    const previousDate = DateTime.now().minus({ days: 1 }).toJSDate();
+    if (!deposit) return;
+    if (!deposit.tokenId || !deposit.token || !deposit.depositDate) throw new Error("Invalid deposit");
+    const previousDate = DateTime.fromISO(deposit.depositDate.toISOString()).minus({ days: 1 }).toJSDate();
```
amateima, line 27: use the deposit date instead of the current date

src/modules/scraper/entry-point/http/controller.ts
```diff
@@ -20,6 +25,17 @@ export class ScraperController {
...
+  @Post("scraper/prices")
+  @ApiTags("scraper")
+  async submitPricesJobs(@Body() body: ProcessPricesBody) {
+    const { fromDepositId, toDepositId } = body;
+    for (let depositId = fromDepositId; depositId <= toDepositId; depositId++) {
+      await this.scraperQueuesService.publishMessage<TokenPriceQueueMessage>(ScraperQueue.TokenPrice, {
+        depositId,
+      });
+    }
+  }
```
amateima, line 37: This is a new endpoint that allows to submit custom messages to the queue that's responsible for getting the token price. This is how we'll update the DB data without re-scraping everything

nicholaspai: AH very nice

PR #18, opened 2022-07-19
feat: implement sticky referral address

amateima:
https://linear.app/uma/issue/ACX-78/make-referrals-sticky


src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -46,7 +47,27 @@ export class DepositReferralConsumer {
...
+      // if no referral address is found, then look for the previous referral address used by this account
+      const previousReferralDeposit = await this.depositRepository.findOne({
+        where: {
+          depositorAddr: deposit.depositorAddr,
+          depositDate: LessThanOrEqual(deposit.depositDate),
+          stickyReferralAddress: Not(IsNull()),
+        },
+        order: { depositDate: { direction: "DESC", nulls: "LAST" } },
+      });
+      stickyReferralAddress = previousReferralDeposit?.stickyReferralAddress;
```
amateima, line 61: if referral address is not found in calldata, then look for the most recent deposit of this wallet address, but older than the current one

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -46,7 +47,27 @@ export class DepositReferralConsumer {
...
+    }
+
+    await this.depositRepository.update(
+      { id: deposit.id },
+      {
+        referralAddress: referralAddress || null,
+        stickyReferralAddress: stickyReferralAddress || referralAddress || null,
```
amateima, line 68: store referral address and sticky referral address as separate fields, so we never lose the on-chain data

nicholaspai, line 68: As noted IRL, I think we should check `referralAddress` first in case user elects to "change" their sticky referral address

nicholaspai (approved):

nicholaspai (approved):

mrice32 (commented):
I've been reading through the architecture a bit, and I was wondering if you could explain how we ensure the ordering is deterministic?

Here are some leading questions I have:

If I have 5 new mainnet blocks with some deposits there and 100 new arbitrum blocks, does the current queuing mechanism ensure that an arbitrum deposit with timestamp 100 is processed before a mainnet deposit with timestamp 99 if they came in at the same time? Note: if we're using the timestamp specified _in_ the deposit, you could conceivably have a deposit in block 100 with a timestamp that is _smaller_ than a deposit in block 99 on the same chain. If we're just processing blocks in order, but are applying stickiness by deposit timestamp, this could cause non determinism.

Similarly, is it deterministic what happens if there are three deposits on three chains with the same deposit timestamp, where two have referral addresses and one does not? Which referral code does the untagged deposit get? Put differently, is there a deterministic order of chains where timestamp 100 on chain A is considered to be earlier than timestamp 100 on chain B, for instance?

I think determinism is very important so we can tell if the system is working correctly or not.

amateima:
> I've been reading through the architecture a bit, and I was wondering if you could explain how we ensure the ordering is deterministic?
> 
> Here are some leading questions I have:
> 
> If I have 5 new mainnet blocks with some deposits there and 100 new arbitrum blocks, does the current queuing mechanism ensure that an arbitrum deposit with timestamp 100 is processed before a mainnet deposit with timestamp 99 if they came in at the same time? Note: if we're using the timestamp specified _in_ the deposit, you could conceivably have a deposit in block 100 with a timestamp that is _smaller_ than a deposit in block 99 on the same chain. If we're just processing blocks in order, but are applying stickiness by deposit timestamp, this could cause non determinism.
> 
> Similarly, is it deterministic what happens if there are three deposits on three chains with the same deposit timestamp, where two have referral addresses and one does not? Which referral code does the untagged deposit get? Put differently, is there a deterministic order of chains where timestamp 100 on chain A is considered to be earlier than timestamp 100 on chain B, for instance?
> 
> I think determinism is very important so we can tell if the system is working correctly or not.

@mrice32 All from above were valid concerns. I modified the PR, so now when a referral address is processed by `DepositReferralConsumer`, all the older referral addressed are checked/updated by the same SQL query. The order in which the deposits are processed is not relevant anymore

>  Similarly, is it deterministic what happens if there are three deposits on three chains with the same deposit timestamp, where two have referral addresses and one does not? Which referral code does the untagged deposit get? Put differently, is there a deterministic order of chains where timestamp 100 on chain A is considered to be earlier than timestamp 100 on chain B, for instance?

I don't know if we can do smth about this and no, I dont have a deterministic order of chains. But the chances of this happening are very low, right? It would require to have the same depositor making deposits at the same deposit time. This can happen only if the deposits are done programmatically and not from the UI

migrations/1659106715328-MaterializedView.ts
```diff
@@ -0,0 +1,48 @@
...
+        select
+          d."depositId",
+          d."depositTxHash",
+          d."sourceChainId",
+          d."destinationChainId",
+          d.amount,
+          t.symbol,
+          t.decimals,
+          d."depositorAddr",
+          d."stickyReferralAddress" as "referralAddress",
```
amateima, line 19: Use the `stickyReferralAddress` column instead of `referralAddress`. `referralAddress` stores only the on-chain referral address

migrations/1659106715329-Deposit.ts
```diff
@@ -0,0 +1,38 @@
...
+    await queryRunner.query(`DROP VIEW IF EXISTS "deposit_referral_stats";`);
+    await queryRunner.query(`
+      CREATE VIEW "deposit_referral_stats" as
+      (
+        select
+          d1.id,
+          count(distinct d2."depositorAddr") as "referralCount",
+          sum((d2.amount / power(10, t.decimals)) * hmp.usd) as "referralVolume"
+        from deposit d1
+        left join deposit d2 on d1."stickyReferralAddress" = d2."stickyReferralAddress"
```
amateima, line 16: this is the same old view, only that `stickyReferralAddress` replaces `referralAddress` column

src/modules/referral/services/queries.ts
```diff
@@ -76,3 +76,27 @@ export const getActiveRefereesCountQuery = () => {
...
+      from deposit d3
+      inner join deposit d4
+        on d3."depositorAddr" = d4."depositorAddr" and d3."depositDate" >= d4."depositDate" and 
+          d4."referralAddress" is not null
+      where d3."depositorAddr" = $1
+      order by d3."depositDate" desc, d4."depositDate" desc
+    ) d1
+    where deposit.id = d1.id and d1."rowNumber" = 1;  
+  `;
+};
```
amateima, line 102: this is the key logic. It updates all the sticky referral addresses for a particular depositor address ($1 param)

src/modules/referral/services/queries.ts
```diff
@@ -76,3 +76,27 @@ export const getActiveRefereesCountQuery = () => {
     where temp.r = 1 and temp."referralAddress" = $1;
   `;
 };
+
+export const updateStickyReferralAddresses = () => {
```
mrice32, line 80: How often does this update get run? On every addition of a new element? Does it run over all previous elements and update them too?

mrice32 (commented):
Just one question!

src/modules/referral/services/queries.ts
amateima: @mrice32 

It runs whenever a new deposit is detected

> Does it run over all previous elements and update them too?

Yes! It performs a join SQL operation to update the referral address of all the deposits performed by the same depositor address

mrice32: Interesting!

Pretty wild that you can do all this in SQL.

mrice32 (approved):
LGTM!

PR #21, opened 2022-07-19
fix: fix capped realized lp fee

amateima:
https://linear.app/uma/issue/ACX-79/user-cannot-see-bridge-fees-and-rewards-on-ui


nicholaspai (approved):

nicholaspai (approved):
Don't you also need to make the change to the SQL query?

PR #26, opened 2022-07-20
fix: check fillTx uniqueness

amateima:


src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
```diff
@@ -45,7 +45,14 @@ export class FillEventsConsumer {
...
 
-    if (deposit.fillTxs.includes({ hash: transactionHash, totalFilledAmount })) return;
+    const fillTxIndex = deposit.fillTxs.findIndex(
+      (fillTx) => fillTx.hash === transactionHash && fillTx.totalFilledAmount === totalFilledAmount,
+    );
+
+    if (fillTxIndex !== -1) {
+      this.logger.warn("Fill event already processed");
+      return;
+    }
```
amateima, line 55: `includes` method is not suitable for checking the existing of an object inside an array. I changed it with `findIndex`

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
```diff
@@ -45,7 +45,14 @@ export class FillEventsConsumer {
       throw new Error("Deposit not found in db");
     }
 
-    if (deposit.fillTxs.includes({ hash: transactionHash, totalFilledAmount })) return;
+    const fillTxIndex = deposit.fillTxs.findIndex(
+      (fillTx) => fillTx.hash === transactionHash && fillTx.totalFilledAmount === totalFilledAmount,
```
nicholaspai, line 49: Isn't transactionHash alone sufficient to identify uniquely the fill?

nicholaspai (approved):
One question

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
nicholaspai: Why wouldn't included work?

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
amateima: I remember you said that we can have multiple fills in a tx

nicholaspai: Ah you're right. So couldn't you then just check totalFilledAmlunt?

amateima: yes, I can. I added the `transactionHash` as being part of the check because I need it anyway to be stored in db

nicholaspai: hm ok

nicholaspai (approved):

PR #27, opened 2022-07-21
fix: add total for referrals table

amateima:
This PR adds the total count for the referrals table data. It is needed to compute the total number of pages

### DIFF
--- a/src/modules/referral/services/queries.ts
+++ b/src/modules/referral/services/queries.ts
@@ -17,6 +17,15 @@ export const getReferralsQuery = () => {
   `;
 };
 
+export const getReferralsTotalQuery = () => {
+  return `
+    select count(*)
+    from deposits_mv as d
+    where d."referralAddress" = $1 or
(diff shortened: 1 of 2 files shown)


mrice32 (approved):
LGTM!

src/modules/referral/services/queries.ts
```diff
@@ -17,6 +17,15 @@ export const getReferralsQuery = () => {
   `;
 };
 
+export const getReferralsTotalQuery = () => {
+  return `
+    select count(*)
+    from deposits_mv as d
+    where d."referralAddress" = $1 or
```
nicholaspai, line 24: So you're saying either a referral has a referral address or a depositor address?

amateima: @nicholaspai 
yes. because you get 25% if the address is the depositor address

amateima: $1 is a particular address that's being checked

PR #29, opened 2022-07-22
improve: change rewards bonus

amateima:
https://linear.app/uma/issue/ACX-100/change-rewards-multiplier-to-2

### DIFF
--- a/migrations/1658221793206-MaterializedView.ts
+++ b/migrations/1658221793206-MaterializedView.ts
@@ -1,0 +1,47 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class MaterializedView1658221793206 implements MigrationInterface {
+  name = "MaterializedView1658221793206";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`drop materialized view if exists deposits_mv`);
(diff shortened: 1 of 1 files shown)


migrations/1658221793206-MaterializedView.ts
```diff
@@ -0,0 +1,47 @@
...
+          when d1."referralCount" >= 20 or d1."referralVolume" >= 500000 then 0.8
+          when d1."referralCount" >= 10 or d1."referralVolume" >= 250000 then 0.7
+          when d1."referralCount" >= 5 or d1."referralVolume" >= 100000 then 0.6
+          when d1."referralCount" >= 3 or d1."referralVolume" >= 50000 then 0.5
+          else 0.4
+        end as "referralRate",
+        case
+            when d."depositDate" < '2022-07-22 17:00:00' then 3
+            else 2
+        end as multiplier
```
amateima, line 32: This new migration file `MaterializedView1658221793206` replaces the definition of the `deposits_mv` materialized view. Everything is identical, except for this `case` where we define the rewards bonus. All deposits before `2022-07-22 17:00:00` will have a multiplier of 3x after which we'll apply a multiplier of 2x

nicholaspai (approved):

PR #30, opened 2022-07-28
improve: use materialized view instead of base tables

amateima:
https://linear.app/uma/issue/ACX-109/get-all-referral-data-from-the-materialized-view

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,8 @@ docker-compose up
 ```shell script
 npm run lint
 ```
+
+## Deployment
+
+- `stage` branch is used to deploy it to the staging environment
+- for production deployment always merge `stage` branch into `master`
--- a/migrations/1658221793207-MaterializedView.ts
+++ b/migrations/1658221793207-MaterializedView.ts
@@ -1,0 +1,48 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class MaterializedView1658221793207 implements MigrationInterface {
+  name = "MaterializedView1658221793207";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`drop materialized view if exists deposits_mv`);
+    await queryRunner.query(`
+      create materialized view deposits_mv as
+        select
+          d."depositId",
+          d."depositTxHash",
+          d."sourceChainId",
+          d."destinationChainId",
+          d.amount,
+          t.symbol,
+          t.decimals,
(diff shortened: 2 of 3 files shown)


migrations/1658221793207-MaterializedView.ts
```diff
@@ -0,0 +1,48 @@
...
+          d."depositTxHash",
+          d."sourceChainId",
+          d."destinationChainId",
+          d.amount,
+          t.symbol,
+          t.decimals,
+          d."depositorAddr",
+          d."referralAddress",
+          d."depositDate",
+          hmp.usd as "tokenUsdPrice",
```
amateima, line 21: add the token usd price as part of the materialized view

src/modules/referral/services/queries.ts
```diff
@@ -47,49 +47,30 @@ export const getTotalReferralRewardsQuery = () => {
...
   return `select count(*) from (
     select distinct on (d."depositorAddr") d."depositorAddr"
-    from deposit d
-    where d."referralAddress" = $1 and
-          d."depositDate" is not null and
-          d."tokenId" is not null and
-          d."priceId" is not null and
-          d.status = 'filled'
+    from deposits_mv as d
+    where d."referralAddress" = $1
```
amateima, line 51: all the queries are basically the same, except that the target table is the `deposits_mv` materialized view

README.md
```diff
@@ -18,3 +18,8 @@ docker-compose up
 ```shell script
 npm run lint
 ```
+
+## Deployment
+
+- `stage` branch is used to deploy it to the staging environment
```
nicholaspai, line 24: +1

migrations/1658221793207-MaterializedView.ts
```diff
@@ -0,0 +1,48 @@
...
+          (d."realizedLpFeePctCapped" / power(10, 18)) * (d.amount / power(10, t.decimals)) * hmp.usd as "realizedLpFeeUsd",
```
nicholaspai, line 22: `realizedLpFeePctCapped` doesn't include `relayerFeePct` yet does it?

nicholaspai (approved):
Left a question

migrations/1658221793207-MaterializedView.ts
amateima: @nicholaspai not yet. The bridge fee re-calculation is coming as a separate PR

PR #31, opened 2022-07-28
chore: disable the materialized view's refresh

amateima:
https://linear.app/uma/issue/ACX-110/disable-the-materialized-views-refresh-while-events-are-re-scraped

This will be merged before re-scraping the blockchain events


nicholaspai (approved):
Makes sense, merge when you're ready

PR #32, opened 2022-07-29
feat: add relayerFee as part of the bridge fee (DO NOT MERGE)

amateima:
https://linear.app/uma/issue/ACX-88/depositrelayerfeepct-should-use-blended-relayerfeepct-of-all-fills

This can't be merged until Evaldo fixes a Zion issue. Currently, the `stage` commits are deployed on both staging and production environments.

### DIFF
--- a/migrations/1659106715325-Deposit.ts
+++ b/migrations/1659106715325-Deposit.ts
@@ -4,12 +4,10 @@ export class Deposit1659106715325 implements MigrationInterface {
   name = "Deposit1659106715325";
 
   public async up(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query(`ALTER TABLE "deposit" ADD "depositRelayerFeePct" numeric NOT NULL DEFAULT '0'`);
     await queryRunner.query(`ALTER TABLE "deposit" ADD "bridgeFeePct" numeric NOT NULL DEFAULT '0'`);
   }
 
   public async down(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(`ALTER TABLE "deposit" DROP COLUMN "bridgeFeePct"`);
-    await queryRunner.query(`ALTER TABLE "deposit" DROP COLUMN "depositRelayerFeePct"`);
   }
 }
--- a/migrations/1659106715325-Deposit.ts
+++ b/migrations/1659106715325-Deposit.ts
@@ -1,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Deposit1659106715325 implements MigrationInterface {
+  name = "Deposit1659106715325";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "deposit" ADD "depositRelayerFeePct" numeric NOT NULL DEFAULT '0'`);
+    await queryRunner.query(`ALTER TABLE "deposit" ADD "bridgeFeePct" numeric NOT NULL DEFAULT '0'`);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "deposit" DROP COLUMN "bridgeFeePct"`);
+    await queryRunner.query(`ALTER TABLE "deposit" DROP COLUMN "depositRelayerFeePct"`);
+  }
+}
--- a/migrations/1659106715326-MaterializedView.ts
+++ b/migrations/1659106715326-MaterializedView.ts
@@ -1,7 +1,7 @@
 import { MigrationInterface, QueryRunner } from "typeorm";
 
-export class MaterializedView1658221793207 implements MigrationInterface {
-  name = "MaterializedView1658221793207";
+export class MaterializedView1659106715326 implements MigrationInterface {
+  name = "MaterializedView1659106715326";
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(`drop materialized view if exists deposits_mv`);
--- a/migrations/1659106715326-MaterializedView.ts
+++ b/migrations/1659106715326-MaterializedView.ts
@@ -4,9 +4,9 @@ export class MaterializedView1659106715326 implements MigrationInterface {
   name = "MaterializedView1659106715326";
 
   public async up(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query(`drop materialized view if exists deposits_mv`);
+    await queryRunner.query(`drop materialized view if exists deposits_mv_v2`);
     await queryRunner.query(`
-      create materialized view deposits_mv as
+      create materialized view deposits_mv_v2 as
         select
           d."depositId",
           d."depositTxHash",
@@ -37,11 +37,11 @@ export class MaterializedView1659106715326 implements MigrationInterface {
         order by d."depositDate" desc;
     `);
     await queryRunner.query(
-      `CREATE UNIQUE INDEX "UK_deposits_mv_depositId_sourceChainId" ON deposits_mv ("depositId", "sourceChainId");`,
+      `CREATE UNIQUE INDEX "UK_deposits_mv_v2_depositId_sourceChainId" ON deposits_mv_v2 ("depositId", "sourceChainId");`,
     );
   }
 
   public async down(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query(`drop materialized view if exists deposits_mv`);
+    await queryRunner.query(`drop materialized view if exists deposits_mv_v2`);
   }
 }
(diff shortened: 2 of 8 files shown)


migrations/1659106715325-Deposit.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Deposit1659106715325 implements MigrationInterface {
+  name = "Deposit1659106715325";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "deposit" ADD "depositRelayerFeePct" numeric NOT NULL DEFAULT '0'`);
+    await queryRunner.query(`ALTER TABLE "deposit" ADD "bridgeFeePct" numeric NOT NULL DEFAULT '0'`);
```
amateima, line 8: add new columns to replace the old `realizedLpFeePctCapped` column

migrations/1659106715326-MaterializedView.ts
```diff
@@ -0,0 +1,47 @@
...
+          d."depositTxHash",
+          d."sourceChainId",
+          d."destinationChainId",
+          d.amount,
+          t.symbol,
+          t.decimals,
+          d."depositorAddr",
+          d."referralAddress",
+          d."depositDate",
+          (d."bridgeFeePct" / power(10, 18)) * (d.amount / power(10, t.decimals)) * hmp.usd as "bridgeFeeUsd",
```
amateima, line 21: `bridgeFeePct ` is used instead of the old `realizedLpFeePctCapped`

src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
```diff
@@ -44,6 +45,7 @@ export class BlocksEventsConsumer {
           status: "pending",
           amount: amount.toString(),
           filled: "0",
+          depositRelayerFeePct: relayerFeePct.toString(),
```
amateima, line 48: store the deposit relayerFeePct in the database

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
```diff
@@ -19,56 +19,91 @@ export class FillEventsConsumer {
...
+    const relayerFeePctChargedToUser = validFills.reduce((cumulativeFee, fill) => {
+      const relayerFee = new BigNumber(fill.fillAmount).multipliedBy(deposit.depositRelayerFeePct);
+      return relayerFee.plus(cumulativeFee);
+    }, new BigNumber(0));
+    const blendedRelayerFeePct = relayerFeePctChargedToUser.dividedBy(deposit.amount).decimalPlaces(0, 1);
+    const bridgeFeePct = blendedRelayerFeePct.plus(fill.realizedLpFeePct);
+    const bridgeFeePctCapped = BigNumber.min(bridgeFeePct, maxBridgeFeePct);
+
+    return bridgeFeePctCapped;
+  }
```
amateima, line 64: this is the important code to be reviewed. This is how the bridge fee is computed

src/modules/scraper/model/deposit.entity.ts
```diff
@@ -15,6 +15,9 @@ export type TransferStatus = "pending" | "filled";
 export type DepositFillTx = {
   hash: string;
   totalFilledAmount: string;
+  fillAmount: string;
+  isSlowRelay: boolean;
+  realizedLpFeePct: string;
```
amateima, line 20: store more info about the fill events in the database

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
```diff
@@ -19,56 +19,91 @@ export class FillEventsConsumer {
...
-    deposit.fillTxs = [...deposit.fillTxs, { hash: transactionHash, totalFilledAmount }];
     deposit.status = new BigNumber(deposit.amount).eq(deposit.filled) ? "filled" : "pending";
+    deposit.bridgeFeePct = bridgeFeePct.toString();
 
     await this.depositRepository.save(deposit);
   }
 
+  private computeBridgeFee(deposit: Deposit, fill: FillEventsQueueMessage) {
+    const maxBridgeFeePct = new BigNumber(10).pow(18).times(0.0012);
+    const validFills = deposit.fillTxs.filter((fill) => !fill.isSlowRelay); // all fills associated with a deposit that are NOT slow fills
```
nicholaspai, line 54: Just a nit but probably nothing to do since this is pretty succinct already:
- You could read `appliedRelayerFeePct` from the FillEvent which is set to 0 for `fill.isSlowRelay == true` and not have to do this `filter` step.

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
```diff
@@ -19,56 +19,91 @@ export class FillEventsConsumer {
...
+    const bridgeFeePct = blendedRelayerFeePct.plus(fill.realizedLpFeePct);
+    const bridgeFeePctCapped = BigNumber.min(bridgeFeePct, maxBridgeFeePct);
+
+    return bridgeFeePctCapped;
+  }
+
+  private fillTxAlreadyProcessed(deposit: Deposit, fill: FillEventsQueueMessage) {
+    const { totalFilledAmount, transactionHash } = fill;
+    const fillTxIndex = deposit.fillTxs.findIndex(
+      (fillTx) => fillTx.hash === transactionHash && fillTx.totalFilledAmount === totalFilledAmount,
```
nicholaspai, line 69: This might not be totally correct. Imagine one transaction hash has multiple fills for _different_ deposits, so both of the fills have the same tx and the same totalFilledAmount.

I would probably check: `depositId`, `originChainId`, AND `totalFilledAmount`

nicholaspai (commented):
Left some questions

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
amateima: > Imagine one transaction hash has multiple fills for different deposits

Wouldn't it emit different events for each deposit?

The `FillEventsConsumer` performs this check of each deposit individually

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
amateima: I like this solution. And in this case I don't need the deposit.relayerFee anymore, right?

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
nicholaspai: Yes, you're right each deposit is a unique event. So i think you're right: for a given deposit the `totalFilledAmount` is unique!

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
nicholaspai: i'm not 100% sure how you use `relayerFee` but yes this is a good solution. It also captures the very edge case (that hasn't happened yet in prod) where a relayer actually [increases their relayer fee](https://github.com/across-protocol/contracts-v2/blob/master/contracts/SpokePool.sol#L413) after the case to further incentivize a potential relayer

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
```diff
@@ -19,56 +19,98 @@ export class FillEventsConsumer {
...
-    deposit.fillTxs = [...deposit.fillTxs, { hash: transactionHash, totalFilledAmount }];
     deposit.status = new BigNumber(deposit.amount).eq(deposit.filled) ? "filled" : "pending";
+    deposit.bridgeFeePct = bridgeFeePct.toString();
 
     await this.depositRepository.save(deposit);
   }
 
+  private computeBridgeFee(deposit: Deposit, fill: FillEventsQueueMessage) {
+    const maxBridgeFeePct = new BigNumber(10).pow(18).times(0.0012);
+    const validFills = deposit.fillTxs.filter((fill) => fill.appliedRelayerFeePct !== "0"); // all fills associated with a deposit that are NOT slow fills
```
nicholaspai, line 61: Does this filter do anything? If the `appliedRelayerFeePct == 0` then its just not added to the following sum right?

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
```diff
@@ -19,56 +19,98 @@ export class FillEventsConsumer {
...
+    const relayerFeePctChargedToUser = validFills.reduce((cumulativeFee, fill) => {
```
nicholaspai, line 62: Should this be renamed `relayerFeeChargedToUser` since its not a percentage?

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
amateima: @nicholaspai 
yes. I did the filtering explicitly to make the code as suggestive as possible

src/modules/scraper/adapter/messaging/FillEventsConsumer.ts
amateima: Good point. I came with this suggestion in the Linear issue, but completely forgot about it

PR #34, opened 2022-07-29
feat: add option to disable events processing

amateima:
Sometimes we may need to stop the fetching of blockchain events (for example during db cleaning operations). Add the option of doing it by using an ENV variable


nicholaspai (approved):
Great idea! LGTM

PR #48, opened 2022-08-16
feat: GET /deposits endpoint

dohaki:
This adds a deposit module that contains an endpoint to retrieve deposits from the db.

TODOs:
- [x] e2e test

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --config ./test/jest-e2e.json",
+    "test:e2e": "jest --maxConcurrency=1 --config ./test/jest-e2e.json",
     "test:e2e:docker": "docker-compose -f docker-compose.e2e.yml up --abort-on-container-exit --exit-code-from backend",
     "typeorm": "node --require ts-node/register ./node_modules/typeorm/cli.js",
     "db:migration:generate": "yarn typeorm -- migration:generate -d ormconfig.ts",
--- a/src/modules/deposit/controller.ts
+++ b/src/modules/deposit/controller.ts
@@ -10,8 +10,8 @@ export class DepositController {
   @Get("deposits")
   @ApiTags("deposits")
   getDeposits(@Query() query: GetDepositsQuery) {
-    const limit = isNaN(parseInt(query.limit)) ? 10 : parseInt(query.limit);
-    const offset = isNaN(parseInt(query.offset)) ? 0 : parseInt(query.offset);
+    const limit = parseInt(query.limit ?? "10");
+    const offset = parseInt(query.offset ?? "0");
     return this.depositService.getDeposits(query.status, limit, offset);
   }
 }
--- a/src/modules/deposit/controller.ts
+++ b/src/modules/deposit/controller.ts
@@ -11,7 +11,7 @@ export class DepositController {
   @ApiTags("deposits")
   getDeposits(@Query() query: GetDepositsQuery) {
     const limit = isNaN(parseInt(query.limit)) ? 10 : parseInt(query.limit);
-    const offset = isNaN(parseInt(query.offset)) ? 10 : parseInt(query.offset);
+    const offset = isNaN(parseInt(query.offset)) ? 0 : parseInt(query.offset);
     return this.depositService.getDeposits(query.status, limit, offset);
   }
 }
--- a/src/modules/deposit/controller.ts
+++ b/src/modules/deposit/controller.ts
@@ -1,0 +1,17 @@
+import { Controller, Get, Query } from "@nestjs/common";
+import { ApiTags } from "@nestjs/swagger";
+import { DepositService } from "./service";
+import { GetDepositsQuery } from "./dto";
+
+@Controller()
+export class DepositController {
+  public constructor(private depositService: DepositService) {}
+
+  @Get("deposits")
+  @ApiTags("deposits")
+  getDeposits(@Query() query: GetDepositsQuery) {
+    const limit = isNaN(parseInt(query.limit)) ? 10 : parseInt(query.limit);
+    const offset = isNaN(parseInt(query.offset)) ? 10 : parseInt(query.offset);
+    return this.depositService.getDeposits(query.status, limit, offset);
+  }
+}
--- a/src/modules/deposit/dto.ts
+++ b/src/modules/deposit/dto.ts
@@ -1,0 +1,27 @@
+import { ApiProperty } from "@nestjs/swagger";
+import { IsEnum, IsNumberString, IsOptional } from "class-validator";
+
+export class GetDepositsQuery {
+  @IsOptional()
+  @IsEnum(
+    {
+      FILLED: "filled",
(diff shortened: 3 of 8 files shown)


src/modules/deposit/service.ts
```diff
@@ -0,0 +1,34 @@
...
+import { InjectRepository } from "@nestjs/typeorm";
+import { Repository } from "typeorm";
+import { Deposit } from "../scraper/model/deposit.entity";
+
+@Injectable()
+export class DepositService {
+  constructor(@InjectRepository(Deposit) private depositRepository: Repository<Deposit>) {}
+
+  public async getDeposits(status = null, limit = 10, offset = 0) {
+    const where = status ? { status } : undefined;
```
amateima, line 11: let's check for the depositDate column to not be `null`

src/modules/deposit/service.ts
```diff
@@ -0,0 +1,34 @@
...
+        skip: offset,
+        order: {
+          depositDate: "desc",
+        },
+      }),
+      this.depositRepository.count({ where }),
+    ]);
+
+    return {
+      deposits,
```
amateima, line 26: I think we need to select only a subset of columns and some of them needs to be renamed/modified in order to match the response expected by the FE

depositDate (from db) -> depositTime (UNIX timestamp)
fillTxs -> the FE is expecting an array of strings, but the DB column is an array of objects.

amateima (changes requested):

src/modules/deposit/service.ts
dohaki: Yea, so currently I am handling the formatting on the client. See https://github.com/across-protocol/frontend-v2/pull/221/files#diff-bbece6ae0b2388ae96ac90b7bdf7db01dee1e8f61e4ff3a008d1cf002b83abc2R36-R49.

I mean these additional columns could be handy at some point? Or is the goal of this API to mirror exactly the structure of the on-chain `Transfers`?

amateima: > I mean these additional columns could be handy at some point?

This is a good question. At this moment we don't have plans to use the additional columns in any way.

src/modules/deposit/service.ts
```diff
@@ -1,14 +1,17 @@
...
 
 @Injectable()
 export class DepositService {
   constructor(@InjectRepository(Deposit) private depositRepository: Repository<Deposit>) {}
 
   public async getDeposits(status = null, limit = 10, offset = 0) {
-    const where = status ? { status } : undefined;
+    const where = {
+      depositDate: Not(IsNull()),
+      status: status || Not(IsNull()),
```
amateima, line 13: This is not a valid TypeORM `or` query. TypeORM allows to add `or` queries by using arrays in the where clause:

```
where: [{ condition1 }, { condition2 }] // where condition1 or condition2
```

Anyway, checking the status column is not needed because it can't be `null`. As you can see [here](https://github.com/across-protocol/scraper-api/blob/stage/src/modules/scraper/model/deposit.entity.ts#L44-L45), it has a default value and it also cannot be `null` because of the NOT NULL restriction coming from [this migration file](https://github.com/across-protocol/scraper-api/blob/stage/migrations/1656106635107-Deposit.ts#L14)

dohaki: So what I try to achieve here is the following
- if `status` param is set, then filter by the given status
- if not, then do not filter by status, i.e. all status that are not null or basically all deposits

I agree that this is not easily understandable. I can rewrite to something like
```js
const baseWhere = { depositDate: Not(IsNull()) };
const where = status ? { ...baseWhere, status } : baseWhere;
```
?

src/modules/deposit/service.ts
dohaki: done

src/modules/deposit/service.ts
amateima: it possible that if status is undefined, then the query will contain no references to the status column

dohaki: yes, you were right 👍 much cleaner now and works

src/modules/deposit/dto.ts
```diff
@@ -0,0 +1,27 @@
...
+    },
+    {
+      message: "Must be one of: 'filled', 'pending'",
+    },
+  )
+  @ApiProperty({ example: "filled", required: false })
+  status: string;
+
+  @IsOptional()
+  @IsNumberString()
```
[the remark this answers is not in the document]
dohaki: Yes makes sense thanks! Added this

package.json
```diff
@@ -17,7 +17,7 @@
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --config ./test/jest-e2e.json",
+    "test:e2e": "jest --maxConcurrency=1 --config ./test/jest-e2e.json",
```
amateima, line 20: I'm just curious, how does this impact our test suite?

test/deposit.e2e-spec.ts
```diff
@@ -0,0 +1,86 @@
+import request from "supertest";
+import { INestApplication, ValidationPipe } from "@nestjs/common";
```
amateima, line 2: Probably it won't impact the tests results, but the app uses [this validation pipe](https://github.com/across-protocol/scraper-api/blob/stage/src/validation.pipe.ts), not the default one. We should replace it in order to fully replicate the production scenario

amateima (changes requested):
It looks great! Just a small nit about the testing file

src/modules/deposit/dto.ts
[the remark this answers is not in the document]
amateima: Didn't know about this option. Nice find

package.json
amateima: Does it solve the issue that is preventing the tests to exit gracefully?

dohaki: > I'm just curious, how does this impact our test suite?

I had some flaky test results when running the tests concurrently. Due to our db cleanup via the `afterAll` hook, basically a race condition. E.g. the e2e tests of the `Scraper Module` would finish, while the `Deposit Module` is still running and delete the deposit fixtures...

> Does it solve the issue that is preventing the tests to exit gracefully?

Unfortunately not... Not quite sure what is preventing a graceful exit tbh. I can dig down a bit as part of this PR or investigate separately?

test/deposit.e2e-spec.ts
dohaki: Ah ok will change

package.json
dohaki: Interesting... if I remove the nested `describe` block, then it actually gracefully shuts down 🤔

src/modules/deposit/controller.ts
```diff
@@ -0,0 +1,17 @@
...
+
+@Controller()
+export class DepositController {
+  public constructor(private depositService: DepositService) {}
+
+  @Get("deposits")
+  @ApiTags("deposits")
+  getDeposits(@Query() query: GetDepositsQuery) {
+    const limit = isNaN(parseInt(query.limit)) ? 10 : parseInt(query.limit);
+    const offset = isNaN(parseInt(query.offset)) ? 0 : parseInt(query.offset);
```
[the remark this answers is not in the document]
dohaki: Yes you are right 🚀

amateima (approved):

PR #49, opened 2022-08-16
fix: improve performance of computing sticky referral addresses

amateima:


src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -49,7 +48,6 @@ export class DepositReferralConsumer {
     }
 
     await this.depositRepository.update({ id: deposit.id }, { referralAddress: referralAddress || null });
-    await this.depositRepository.query(updateStickyReferralAddresses(), [deposit.depositorAddr]);
```
amateima, line 52: instead of running this query for every new deposit (which can be very inefficient), I created a new cron job which runs 1 minute before the cron that refreshes the materialized view. This way we'll run the query a single time and not for each new deposit

src/modules/referral/services/queries.ts
```diff
@@ -94,7 +94,6 @@ export const updateStickyReferralAddresses = () => {
       inner join deposit d4
         on d3."depositorAddr" = d4."depositorAddr" and d3."depositDate" >= d4."depositDate" and 
           d4."referralAddress" is not null
-      where d3."depositorAddr" = $1
```
amateima, line 97: update all sticky referral addresses, not only for a particular address. This query will run inside a cron job every 5 minutes

src/modules/referral/services/cron-service.ts
```diff
@@ -12,11 +13,11 @@ export class ReferralCronService {
   constructor(
     @InjectRepository(Deposit) private depositRepository: Repository<Deposit>,
     private appConfig: AppConfig,
-  ) {}
+  ) { }
```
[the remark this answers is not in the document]
amateima: yes

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
nicholaspai: so this means sticky referral addresses could be out of sync for up to 1 minute?

amateima: no, because when the MATERIALIZED VIEW is updated, the sticky addresses are already populated. 

So the flow is:
- using the processing queues, the deposit table is updated constantly with data from events and other sources
- every 5 minutes (starting with minute 0: 0, 5, 10, etc) the sticky addresses are updated
- every 5 minutes (starting with minute 1: 1, 6, 11, etc) the MATERIALIZED VIEW (which the referral data is served from) is updated

nicholaspai: What is the current flow? With the new flow, does this mean that if the user is expecting a sticky referral to change and then sends a deposit before the materialized view is updated after 5 mins, then the sticky referral address wouldn't have changed? I expect this to be a rare case regardless

PR #50, opened 2022-08-17
feat: add e2e test

amateima:
https://linear.app/uma/issue/ACX-20/deposits-should-not-be-duplicated-in-the-db

This PR adds an e2e test to validate that the same deposit can't be inserted multiple times in the database. The deposits are unique by deposit id and source chain id.


docker-compose.e2e.yml
```diff
@@ -14,7 +14,7 @@ services:
       - 9229:9229
     command: ./docker/e2e.command.sh
     env_file:
-      - .env
+      - .env.test
```
amateima, line 17: use a separate env file for the testing infrastructure

src/utils.ts
```diff
@@ -1,4 +1,13 @@
...
 export const wait = (seconds = 1) =>
   new Promise((res) => {
     setTimeout(res, 1000 * seconds);
   });
+
+export const EnhancedCron = (cronExpression: string) => {
+  if (process.env.DISABLE_CRONS != "true") return applyDecorators(Cron(cronExpression));
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  else return () => {};
+};
```
amateima, line 13: a new decorator which will be used instead of the default @Cron. This will check automatically if the cron jobs have been disabled

src/modules/scraper/adapter/db/deposit-fixture.ts
```diff
@@ -0,0 +1,45 @@
+import { Injectable } from "@nestjs/common";
+import { InjectRepository } from "@nestjs/typeorm";
+import { Repository } from "typeorm";
+import { Deposit } from "../../model/deposit.entity";
+
+@Injectable()
+export class DepositFixture {
```
amateima, line 7: this is a class used to perform database operations during e2e tests. Each db model will have its own fixture file

test/scraper.e2e-spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    app = moduleFixture.createNestApplication();
+    await app.init();
+  });
+
+  it("should not store the same deposit twice times", async () => {
+    await app.get(DepositFixture).insertDeposit({ depositId: 1, sourceChainId: 1, destinationChainId: 10 });
+    try {
+      await app.get(DepositFixture).insertDeposit({ depositId: 1, sourceChainId: 1, destinationChainId: 10 });
+      expect(true).toEqual(false);
```
amateima, line 23: this line should not be reached ever

test/scraper.e2e-spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    } catch (error) {
+      expect(error instanceof QueryFailedError).toEqual(true);
```
amateima, line 25: assert that the second insertion was not made and an error was thrown

test/scraper.e2e-spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+  });
+
+  afterAll(async () => {
+    await app.get(DepositFixture).deleteAllDeposits();
+    await app.close();
+  });
```
amateima, line 32: clean up everything after tests are finished

test/scraper.e2e-spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+describe("Scraper module", () => {
+  let app: INestApplication;
+
+  beforeAll(async () => {
+    const moduleFixture = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
```
dohaki, line 15: This app instance can then be used to test endpoints with things like [supertest](https://github.com/visionmedia/supertest) ?

dohaki (approved):
LGTM 👍  Left one small question and a suggestion

test/scraper.e2e-spec.ts
amateima: Exactly! It'd be smth like

```js
import request from "supertest";

const response = await request(app.getHttpServer()).get("url");
expect(response.data)...
expect(response.statusCode)...
```

[NestJS docs example](https://docs.nestjs.com/fundamentals/testing#end-to-end-testing)

test/scraper.e2e-spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it("should not store the same deposit twice times", async () => {
+    await app.get(DepositFixture).insertDeposit({ depositId: 1, sourceChainId: 1, destinationChainId: 10 });
+    try {
+      await app.get(DepositFixture).insertDeposit({ depositId: 1, sourceChainId: 1, destinationChainId: 10 });
+      expect(true).toEqual(false);
+    } catch (error) {
+      expect(error instanceof QueryFailedError).toEqual(true);
+    }
```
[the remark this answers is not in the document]
amateima: good suggestion @dohaki . thx 🙌

PR #62, opened 2022-08-30
fix: increase crons time difference

amateima:


dohaki (approved):

src/modules/referral/services/cron-service.ts
```diff
@@ -13,11 +13,12 @@ export class ReferralCronService {
   constructor(
     @InjectRepository(Deposit) private depositRepository: Repository<Deposit>,
     private appConfig: AppConfig,
-  ) { }
+  ) {}
 
   // run cron every 5 minutes
-  @EnhancedCron("0 1-59/5 * * * *")
+  @EnhancedCron("0 2-59/5 * * * *")
```
mrice32, line 19: What does this change do? It looks like it just changes the offset?

amateima: Yes. More details [here](https://github.com/across-protocol/scraper-api/pull/61#issue-1354531134)

PR #69, opened 2022-09-02
fix: use depositDate instead of createdDate

dohaki:


src/modules/deposit/service.ts
```diff
@@ -39,7 +39,7 @@ export class DepositService {
 export function formatDeposit(deposit: Deposit) {
   return {
     depositId: deposit.depositId,
-    depositTime: Math.round(DateTime.fromISO(deposit.createdAt.toISOString()).toSeconds()),
+    depositTime: Math.round(DateTime.fromISO(deposit.depositDate.toISOString()).toSeconds()),
```
amateima, line 42: The luxon interface is so easy to easy

amateima (approved):

PR #71, opened 2022-09-06
fix: improve sticky referral cron

amateima:
This PR improves the way sticky referral addresses are updated and also provides two strategies for doing it configurable by an environment variable, so we can switch between them whenever we want. 

Please follow my comments in the `Files changed` tab. 

🕹This will be tested firstly on staging


src/modules/referral/services/cron-service.ts
```diff
@@ -3,43 +3,64 @@ import { Repository } from "typeorm";
...
-import { EnhancedCron } from "../../../utils";
-import { updateStickyReferralAddresses } from "../services/queries";
+import { EnhancedCron, wait } from "../../../utils";
+import { updateStickyReferralAddressesQuery } from "../services/queries";
+import { StickyReferralAddressesMechanism } from "src/modules/configuration";
 
 @Injectable()
 export class ReferralCronService {
   private logger = new Logger(ReferralCronService.name);
+  private semaphore = false;
```
amateima, line 13: This is unlikely to happen, but in case the previous execution of the cron doesn't finish, don't trigger a new one.

src/modules/referral/services/queries.ts
```diff
@@ -99,3 +99,26 @@ export const updateStickyReferralAddresses = () => {
...
+            from deposit d2
+            where d2."depositorAddr" = d1."depositorAddr" and
+                  d2."depositDate" <= d1."depositDate" and
+                  d2."referralAddress" is not null
+            order by d2."depositDate" desc
+            limit 1
+            ) d3 on d1."depositorAddr" = d3."depositorAddr"
+        where d1."depositorAddr" = $1
+    ) d4
+    where deposit.id = d4.id;
```
amateima, line 122: Query to update the sticky referrals for a particular depositor address only

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -48,6 +50,16 @@ export class DepositReferralConsumer {
...
+
+    if (this.appConfig.values.stickyReferralAddressesMechanism === StickyReferralAddressesMechanism.Queue) {
+      const hasDepositsWithReferrals = await this.depositRepository.findOne({
+        where: { depositorAddr: deposit.depositorAddr, referralAddress: Not(IsNull()) },
+      });
+
+      if (hasDepositsWithReferrals) {
+        await this.depositRepository.query(updateStickyReferralAddressesForDepositor(), [deposit.depositorAddr]);
+      }
+    }
```
amateima, line 62: If the sticky referrals are not updated by the cron, run the query which updates the referrals for this depositor only. If the depositor didn't use a referrer address at all, then don't run the query

src/modules/referral/services/cron-service.ts
```diff
@@ -3,43 +3,64 @@ import { Repository } from "typeorm";
...
-    this.logger.log("start refreshMaterializedViewCron()");
-    if (this.appConfig.values.enableReferralsMaterializedViewRefresh) {
+  @EnhancedCron("0 */10 * * * *")
+  async startCrons() {
+    if (this.semaphore === true) {
+      return;
+    }
+    this.semaphore = true;
+
+    await this.updateStickyReferralAddresses();
```
amateima, line 27: Sticky referrals updated by a cron have more sense when the database is rebuilt from scratch and a huge number of events are fetched. This would put a lot of pressure on the `DepositReferralConsumer` which updates sticky referrals on each new deposit.

.env.sample
```diff
@@ -15,4 +15,5 @@ WEB3_NODE_URL_137=https://infura.io
 REFERRAL_DELIMITER_START_TIMESTAMP=1657290720
 ENABLE_SPOKE_POOLS_EVENTS_PROCESSING=true
 ENABLE_REFERRALS_MATERIALIZED_VIEW_REFRESH=true
+STICKY_REFERRAL_ADDRESSES_MECHANISM=queue
```
amateima, line 18: `queue | cron | disabled` are the possible values here

nicholaspai: do you think its worth commenting that these are possible values?

migrations/1662485183051-Deposit.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Deposit1662485183051 implements MigrationInterface {
+  name = "Deposit1662485183051";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`CREATE INDEX "IX_deposit_depositorAddr" ON "deposit" ("depositorAddr") `);
```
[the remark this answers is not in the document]
nicholaspai: How does this improve speed so much?

src/modules/referral/services/cron-service.ts
```diff
@@ -3,43 +3,64 @@ import { Repository } from "typeorm";
...
   ) {}
 
-  // run cron every 8 minutes
-  @EnhancedCron("0 4-59/8 * * * *")
-  async refreshMaterializedViewCron() {
-    this.logger.log("start refreshMaterializedViewCron()");
-    if (this.appConfig.values.enableReferralsMaterializedViewRefresh) {
+  @EnhancedCron("0 */10 * * * *")
+  async startCrons() {
+    if (this.semaphore === true) {
```
nicholaspai, line 22: this could be condensed to one line right? `if (this.semaphore) return`

.env.sample
amateima: Good point. Added comment [here](https://github.com/across-protocol/scraper-api/commit/30c6413d3ddef4a967bc4ed4e6a7206f1cc18538)

src/modules/referral/services/cron-service.ts
nicholaspai: So we relieve the pressure by waiting for a 30s cooldown?

nicholaspai: If so, then the cooldown length is important and we should consider making the length configurable by env var

migrations/1662485183051-Deposit.ts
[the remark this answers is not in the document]
amateima: That query fetches records using a WHERE clause which filters the deposits by `depositorAddr` field. Before adding this INDEX, Postgres had to perform a full table scan in order to extract the deposits of a particular depositor address. 
This SQL INDEX adds all `depositorAddr` in a quick lookup table which is queried very fast.

src/modules/referral/services/cron-service.ts
amateima: Yes

nicholaspai (commented):
Left comments

src/modules/referral/services/cron-service.ts
amateima: The cooldown is more of an experiment to see if it produces any effects. Usually, after an intensive SQL query is executed (as the update of sticky referrals), Postgres performs some clean-up operations (delete temp files, perform `VACUUM` operation), but these are meant to run in the background and usually they don't impact the other queries. They are more as a maintenance process

migrations/1662485183051-Deposit.ts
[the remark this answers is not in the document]
mrice32: Great idea!!

src/modules/referral/services/cron-service.ts
mrice32: Just out of curiosity, what bad things would this cause if two ran at the same time?

Just thinking about the other extreme, where one run hangs because some promise never returned, and then the cron just silently stops. I don't think this is _likely_ but it seems possible at least.

src/modules/referral/services/cron-service.ts
mrice32: It would be nice if there was a call you could make to postgres asking it to cleanup and telling you when done.

src/modules/referral/services/cron-service.ts
```diff
@@ -3,43 +3,62 @@ import { Repository } from "typeorm";
...
     private appConfig: AppConfig,
   ) {}
 
-  // run cron every 8 minutes
-  @EnhancedCron("0 4-59/8 * * * *")
-  async refreshMaterializedViewCron() {
-    this.logger.log("start refreshMaterializedViewCron()");
-    if (this.appConfig.values.enableReferralsMaterializedViewRefresh) {
+  @EnhancedCron("0 */10 * * * *")
+  async startCrons() {
```
mrice32, line 21: Is it possible something in this method throws and then the semaphore never gets reset?

src/modules/referral/services/cron-service.ts
```diff
@@ -3,43 +3,62 @@ import { Repository } from "typeorm";
...
-    try {
-      await this.depositRepository.query(updateStickyReferralAddresses());
-    } catch (error) {
-      this.logger.error(error);
+  private async updateStickyReferralAddresses() {
+    this.logger.log("start updateStickyReferralAddresses()");
+
+    // sticky referral addresses are updated either by this cron or by the DepositReferralConsumer
+    if (this.appConfig.values.stickyReferralAddressesMechanism !== StickyReferralAddressesMechanism.Cron) {
+      this.logger.log(`disabled updateStickyReferralAddresses()`);
```
mrice32, line 53: Should we specify that updateStickyReferralAddresses isn't necessarily disabled, just the cron mechanism to call it?

mrice32 (commented):
Looks awesome, only minor comments!

dohaki (approved):
🚀  looking forward to get this merged

src/modules/referral/services/cron-service.ts
amateima: > Just out of curiosity, what bad things would this cause if two ran at the same time?

Firstly, the tables impacted by these 2 intensive queries will become very slow and most probably they won't finish until future execution are triggered. This will put pressure on the DB. In time the resources usage will start to increase significantly (cpu and memory up to 100%)

src/modules/referral/services/cron-service.ts
amateima: Great observation 👍 
No, theoretically this method should not throw errors, but I wrapped everything in a `try catch finally` block. This will make it less error prone in case of future changes. 

New version ⬇️

https://github.com/across-protocol/scraper-api/blob/109e9c4ff4c641687e3ae8eade4cf88562f2bd73/src/modules/referral/services/cron-service.ts#L22-L34

src/modules/referral/services/cron-service.ts
amateima: Yes. I updated the log message

PR #73, opened 2022-09-08
fix: improve sticky referral cron

amateima:
This is the PR for deploying into production the PR #71.

It has been tested on staging with the following DB usage. 

Screenshot 2022-09-09 at 00 57 07
Screenshot 2022-09-09 at 00 56 51

The spikes are caused by the refresh of the materialized view which holds the referral rewards data.


dohaki (approved):
Nice an improvement indeed 👍

PR #75, opened 2022-09-19
feat: add discord auth endpoint

amateima:
This PR adds Discord authentication using OAuth2 protocol and JWT based authorization for authenticated requests alongside a new User entity

The Discord authentication process is described [here](https://github.com/across-protocol/scraper-api/blob/amatei/acx-259-add-discord-auth-endpoint/docs/oauth.md)

cc @james-a-morris


amateima:
The Discord authentication process is described [here](docs/oauth.md)

cc @james-a-morris

mrice32:
> The Discord authentication process is described [here](https://github.com/across-protocol/scraper-api/blob/amatei/acx-259-add-discord-auth-endpoint/docs/oauth.md)

This is awesome @amateima!

docs/oauth.md
```diff
@@ -0,0 +1,31 @@
+# OAuth2
+
+## Login with Discord using OAuth2 protocol
+
+### Instructions
+
+1. Initiate the login process by accessing the following URL from the frontend application:
```
mrice32, line 7: What do you mean by access? Is this a redirect?

src/modules/user/services/user.service.ts
```diff
@@ -0,0 +1,44 @@
...
+@Injectable()
+export class UserService {
+  constructor(@InjectRepository(User) private userRepository: Repository<User>) {}
+
+  public async createUserFromDiscordId({ discordId }: { discordId: string }) {
+    let user = await this.userRepository.findOne({ where: { discordId } });
+
+    if (!user) {
+      const shortId = await this.createShortId();
+      const uuid = this.createUUID();
```
mrice32, line 18: Why do we have 3 ids (the primary key, the short id and the uuid) that are all generated by our application? What are they used for?

.env.sample
```diff
@@ -21,3 +21,7 @@ DISABLE_CRONS=false
 # Following distances in blocks to guarantee finality on each chain. Format should be a map chainId -> blocks.
 # E.g. { "137": 100 }
 FOLLOWING_DISTANCES=
+JWT_SECRET=secret
+DISCORD_CLIENT_ID=
+DISCORD_CLIENT_SECRET=
+DISCORD_REDIRECT_URI=
```
mrice32, line 27: Similar to other fields, it might be useful to either provide a comment or sample value for these.

mrice32 (commented):
This is really cool! Only a few minor comments and questions.

src/modules/auth/auth.module.ts
```diff
@@ -0,0 +1,27 @@
...
+@Module({
+  providers: [DiscordStrategy, AuthService],
+  controllers: [AuthController],
+  imports: [
+    PassportModule.register({}),
+    HttpModule,
+    UserModule,
+    JwtModule.register({
+      secret: configValues.auth.jwtSecret,
+      signOptions: { expiresIn: "7d" },
```
dohaki, line 21: Have we also considered using a refresh token pattern in addition to the access token we issue? It could be a bit over-engineered but I was wondering if you have thought about that.

dohaki (approved):
Looks good 👍 I left only one question regarding refresh tokens

docs/oauth.md
amateima: yes. The login button would be actually a link

```html
<a href="<DISCORD_OAUTH_URL>">Login</button>
```

src/modules/user/services/user.service.ts
amateima: For now only the `id` is used, but I added the `uuid` and the `shortId` in case we need to identify and expose users by non-incremental and random identifiers.

src/modules/auth/auth.module.ts
amateima: Didn't think about them in particular, but yeah, one disadvantage of the current implementation is the fact that the user will be logged out from the FE app when the token will expire. I can create a separate issue for it

docs/oauth.md
mrice32: Makes sense!

PR #78, opened 2022-09-22
build: fix docker compose arm and add ci

dohaki:
This allows to run the docker setup on ARM-based machines.


.github/workflows/lint-build-test.yaml
```diff
@@ -0,0 +1,16 @@
+name: Lint, build and test
+
+on:
+  push:
+    branches:
+      - stage
```
amateima, line 6: shouldn't master be added here as well? The base branch can change again after the PR has been made 🤔

Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM node:16-stretch AS development
+FROM node:16-alpine AS development
+
+RUN apk --no-cache add --virtual .builds-deps build-base python3 git
```
amateima, line 3: I'm curious why stretch needs to be replaced with alpine. Is this related to the ARM issues you had?

docker-compose.yml
```diff
@@ -33,23 +31,17 @@ services:
...
+    command: ["redis-server", "--appendonly", "yes"]
     environment:
       REDIS_PASSWORD: ${REDIS_PASSWORD}
-    networks:
-      - webnet
-    volumes: 
+    volumes:
       - scraper-api-redis-volume:/data
-networks:
-  webnet:
```
amateima, line 52: Did you remove the network in order to allow the DB to be accessed by the BE app started outside `docker-compose`?

.github/workflows/lint-build-test.yaml
```diff
@@ -0,0 +1,16 @@
+name: Lint, build and test
```
amateima, line 1: I see the lint and build commands, but no testing command

.github/workflows/lint-build-test.yaml
dohaki: Yea I can also add `master` here. But if you look at the line below, you can see `pull_request`. That means that on every PR for every base this workflow will run. In addition, we run on pushing / merging to `stage` via this `push` trigger.

Dockerfile
dohaki: Actually, it doesn't have to necessarily. The most important line is
```
RUN apk --no-cache add --virtual .builds-deps build-base python3 git
```

With stretch, I would need to change this to the `apt` equivalent I guess. I prefer `alpine` though because the image is way smaller (~ 5 MB) in contrast to stretch (~ 40 MB). Or is there any specific reason to use the `stretch` variant?

docker-compose.yml
dohaki: I had `ECONNREFUSED` errors in the BE container when started via `docker-compose`. I could fix it by removing this and using the default approach as described here https://docs.docker.com/compose/networking/.

.github/workflows/lint-build-test.yaml
dohaki: Yea I played around with running the e2e tests as part of the pipeline but got stuck. I wanted to finish up the other PR first before continuing on that.

amateima (approved):

.github/workflows/lint-build-test.yaml
dohaki: Ah ok yea then I will add master here. makes sense

PR #79, opened 2022-09-22
feat: POST and PATCH `/users/wallets` endpoints

dohaki:
Adds `POST /users/wallets` and `PATCH /users/wallets` endpoints. Also adds e2e tests for the user module.

Note: Base is set to https://github.com/across-protocol/scraper-api/pull/78 for testability.


src/modules/user/model/wallet.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+export class Wallet {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column({ unique: true })
+  userId: number;
+
+  @OneToOne(() => User)
+  @JoinColumn([{ name: "userId", referencedColumnName: "id" }])
+  user?: User;
```
dohaki, line 24: I think we also need to add some reference in the `user.entity.ts` right @amateima? Or `GET /users/wallets` in order for the FE to be able to know the linked wallet

src/modules/user/services/wallet.service.ts
```diff
@@ -0,0 +1,85 @@
...
+    const user = await this.userRepository.findOne({
+      where: {
+        id: userId,
+      },
+    });
+
+    if (!user) {
+      throw new UserNotFoundException();
+    }
+  }
```
amateima, line 84: I think we remove this method and use the existing:
https://github.com/across-protocol/scraper-api/blob/847f4304a754a16355d44cc075f5f9257dd23a76/src/modules/user/services/user.service.ts#L26

test/fixtures/user.ts
```diff
@@ -0,0 +1,13 @@
+import { User } from "../../src/modules/user/model/user.entity";
+
+export function mockUserEntity(overrides?: Partial<User>): User {
+  return {
+    id: Math.floor(Date.now() / 1000),
```
amateima, line 5: Is it necessary to mock the `id` parameter? It should be created by the database automatically

src/modules/user/services/wallet.service.ts
```diff
@@ -0,0 +1,85 @@
...
+        userId,
+        walletAddress,
+      });
+    } else {
+      wallet.walletAddress = walletAddress;
+    }
+
+    wallet = await this.walletRepository.save(wallet);
+
+    return wallet;
```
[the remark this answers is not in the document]
amateima: Yes, good point. Looks like it works, but needs to be tested

test/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { sign, SignOptions } from "jsonwebtoken";
+import { configValues } from "../src/modules/configuration/index";
+
+export function generateJwtForUser(user: { id: number; shortId: string; uuid: string }, jwtOptions?: SignOptions) {
+  const payload = { id: user.id, uuid: user.uuid, shortId: user.shortId };
+  return sign(payload, configValues().auth.jwtSecret, jwtOptions);
+}
```
amateima, line 7: Can we remove this method and use the existing 
https://github.com/across-protocol/scraper-api/blob/847f4304a754a16355d44cc075f5f9257dd23a76/src/modules/auth/services/auth.service.ts#L9
?

src/modules/user/entry-points/http/user.controller.ts
```diff
@@ -16,4 +19,54 @@ export class UserController {
...
+    await this.walletService.assertUserExists(userId);
+
+    const wallet = await this.walletService.upsertWallet({
+      userId,
+      walletAddress,
+    });
+
+    return {
+      wallet,
+    };
```
amateima, line 45: I feel like we need to keep the controllers structure as simple as possible. Ideally a particular HTTP handler would make a single call to a particular service. In our case it'd be something like `this.walletService.linkUserWallet(...)`

src/modules/user/model/wallet.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+} from "typeorm";
+import { User } from "./user.entity";
+
+@Entity()
+@Unique("UK_wallet_id_walletAddress", ["id", "walletAddress"])
+export class Wallet {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column({ unique: true })
```
amateima, line 19: Can we use the same approach as we used for the `UK_wallet_id_walletAddress` constraint and set it as a decorator attached to the `Wallet` class? It's nice to have a single place to look for indexes and constraints

migrations/1663773308937-Wallet.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Wallet1663773308937 implements MigrationInterface {
+  name = "Wallet1663773308937";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "wallet" ("id" SERIAL NOT NULL, "userId" integer NOT NULL, "walletAddress" character varying NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), "updatedAt" TIMESTAMP NOT NULL DEFAULT now(), CONSTRAINT "UQ_35472b1fe48b6330cd349709564" UNIQUE ("userId"), CONSTRAINT "UK_wallet_id_walletAddress" UNIQUE ("id", "walletAddress"), CONSTRAINT "REL_35472b1fe48b6330cd34970956" UNIQUE ("userId"), CONSTRAINT "PK_bec464dd8d54c39c54fd32e2334" PRIMARY KEY ("id"))`,
```
amateima, line 8: I don't know, maybe it's just me, but the Scraper API contains some DB models that are fundamental for blockchains:  `block`, `transaction`. In our case `wallet` is not any blockchain wallet, but a specific wallet linked by a user. What do you think if we rename this model to `UserWallet` (which translates to `user_wallet` table name)?

dohaki: Yes makes sense. I actually first named it like that 😅

test/fixtures/user.ts
dohaki: Yea the id gets overridden by the db but sometimes I need to mock the whole `User` entity without saving it in the db. For example for test cases with non-existing users.

test/utils.ts
dohaki: I first tried to use this service in the tests like
```ts
const jwt = app.get(AuthService).generateJwtForUser(user)
```
but then I got an error
```
secretOrPrivateKey must have a value
```
and the service that I got looked like
```
AuthService {
  jwtService: JwtService {
    options: { secret: undefined, signOptions: [Object] },
    logger: Logger { context: 'JwtService', options: {} }
  }
}
```

dohaki: any idea why that is?

src/modules/user/entry-points/http/dto.ts
```diff
@@ -0,0 +1,16 @@
...
+  @ApiProperty()
+  signature: string;
+
+  @IsEthereumAddress()
+  @ApiProperty()
+  walletAddress: string;
+
+  @IsString()
+  @ApiProperty()
+  discordId: string;
```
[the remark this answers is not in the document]
dohaki: It actually checks for existence also. So if the `signature` is missing in the body, then it returns a 400 with 
```
{"error":"ValidationError","message":"Please check the data","errors":[{"property":"signature","children":[],"constraints":{"isString":"signature must be a string"}}]}
```

src/modules/user/services/wallet.service.ts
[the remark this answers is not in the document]
dohaki: I changed it to use the native `upsert` method of the [TypeORM Repository API](https://typeorm.io/repository-api#repository-api). AFAICT it builds up the required atomic query like you suggested @james-a-morris.

src/modules/user/entry-points/http/user.controller.ts
```diff
@@ -16,4 +19,54 @@ export class UserController {
       user,
     };
   }
+
+  @Post("wallets")
```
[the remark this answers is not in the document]
dohaki: done

src/modules/user/services/wallet.service.ts
dohaki: done

src/modules/user/entry-points/http/user.controller.ts
dohaki: done

src/modules/user/model/wallet.entity.ts
dohaki: done

migrations/1663773308937-Wallet.ts
dohaki: done

test/utils.ts
amateima: I'll take a look and come back with an answer

amateima (commented):
Looks awesome. Taking a look now on the `JwtService` issue

src/modules/user/services/user-wallet.service.ts
```diff
@@ -0,0 +1,104 @@
...
+    walletAddress,
+    signature,
+    discordId,
+  }: {
+    userId: number;
+    walletAddress: string;
+    signature: string;
+    discordId: string;
+  }) {
+    this.verifySignedDiscordId({
```
amateima, line 28: I think it's good to introduce the following assertion at the beginning of both POST and PATCH handler

```
1. select wallet with walletAddress
2. if wallet.userId === userId then do nothing and return data directly
3. if wallet.userId !== userId throw a BadRequest error
```

src/modules/user/entry-points/http/user.controller.ts
```diff
@@ -16,4 +19,38 @@ export class UserController {
       user,
     };
   }
+
```
amateima, line 22: I know you asked me this before, but let's add a `@Get("me/wallets")` handler here and the response body can be like
```
{
  walletAddress: "<USER_WALLET_ADDRESS_HERE>"
}
```

This can also be the response of the POST and PATCH requests

amateima (changes requested):

src/modules/user/model/user-wallet.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+  PrimaryGeneratedColumn,
+  UpdateDateColumn,
+  OneToOne,
+  JoinColumn,
+  Unique,
+} from "typeorm";
+import { User } from "./user.entity";
+
+@Entity()
+@Unique("UK_userWallet_id_walletAddress", ["id", "walletAddress"])
```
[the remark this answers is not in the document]
dohaki: yea you are right. changed it

src/modules/user/services/user-wallet.service.ts
```diff
@@ -0,0 +1,104 @@
...
+      signature,
+      discordIdMessage: discordId,
+      walletAddress,
+    });
+
+    await this.userService.getUserByAttributes({ id: userId }, true);
+
```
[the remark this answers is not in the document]
dohaki: good catch! I added a test case for this

src/modules/user/services/user-wallet.service.ts
```diff
@@ -0,0 +1,104 @@
...
+    walletAddress,
+    signature,
+    discordId,
+  }: {
+    userId: number;
+    walletAddress: string;
+    signature: string;
+    discordId: string;
+  }) {
+    await this.assertWalletForUserExists(userId);
```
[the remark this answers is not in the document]
dohaki: ok I removed this assertion. post and patch behave the same now bascially

src/modules/user/services/user-wallet.service.ts
dohaki: done

src/modules/user/entry-points/http/user.controller.ts
dohaki: done

amateima (approved):

PR #83, opened 2022-09-26
ci: add e2e tests to ci

dohaki:
This adds running the e2e tests as part of the CI.


.env.test
```diff
@@ -0,0 +1,29 @@
...
+DISABLE_CRONS=true
+# Following distances in blocks to guarantee finality on each chain. Format should be a map chainId -> blocks.
+# E.g. { "137": 100 }
+FOLLOWING_DISTANCES=
+JWT_SECRET=secret
+# DISCORD_CLIENT_ID and DISCORD_CLIENT_SECRET are the credentials of the Discord app and they are obtained from https://discord.com/developers/applications.
+DISCORD_CLIENT_ID=clientId
+DISCORD_CLIENT_SECRET=clientSecret
+# the url accessed after the Discord authorization processed is fulfilled
+DISCORD_REDIRECT_URI=http://localhost
```
dohaki, line 29: I decided to check this file into the repository because it doesn't contain any sensitive information and allows anyone to checkout the repo and run the e2e tests directly locally. Makes sense to you @amateima?

docker/e2e.command.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
 
-yarn install
 yarn db:migration:run
-yarn test:e2e
+yarn test:e2e --force-exit
```
dohaki, line 4: I guess this is the only workaround for now until we find the real issue for `jest` not exiting correctly.

src/modules/user/adapter/db/user-fixture.ts
```diff
@@ -14,13 +14,16 @@ export class UserFixture {
...
     return {
-      discordAvatar: "avatar",
+      id: 0,
+      discordAvatar: "https://picsum.photos/200",
       discordId: "1",
       discordName: "name",
       shortId: this.createShortId(),
       uuid: this.createUUID(),
+      createdAt: new Date(),
+      updatedAt: new Date(),
```
amateima, line 26: I feel that the id, createdAt and updatedAt shouldn't be given here by default, but let the database to populate them. If custom values need to be provided for these fields, then the `overrides` function parameter should be used

test/airdrop.e2e-spec.ts
```diff
@@ -126,8 +127,7 @@ describe("POST /airdrop/upload/rewards", () => {
   });
 
   it("should not be authorized if not admin", async () => {
-    const userJwt = app.get(JwtService).sign({ roles: [Role.User] });
-    console.log({ userJwt });
+    const userJwt = app.get(JwtService).sign({ roles: [Role.User] }, { secret: configValues().auth.jwtSecret });
```
amateima, line 130: hmm, why is the secret needed to be passed explicitly here? It is already picked up from [here](https://github.com/across-protocol/scraper-api/blob/stage/src/modules/auth/auth.module.ts#L21-L24)

amateima (changes requested):

src/modules/user/adapter/db/user-fixture.ts
dohaki: I initially added them due to some type incompatibilities when mocking a non-existing user entity. Because this user entity is not added to the db, I couldn't cast it to the type `User`. But I guess it doesn't have to be. I changed it [here](https://github.com/across-protocol/scraper-api/pull/83/commits/682334bb81ff20424aefa8494ec88c8bf8806f16).

test/airdrop.e2e-spec.ts
dohaki: Yea this is a bit weird. I couldn't figure out why. So if you run the dockerized e2e tests via `yarn test:e2e:docker` then the service picks up the correct secret from the env vars. But if I run the tests in a standalone way (also in the CI) then this test fails with
```
    secretOrPrivateKey must have a value

      128 |
      129 |   it("should not be authorized if not admin", async () => {
    > 130 |     const userJwt = app.get(JwtService).sign({ roles: [Role.User] });
          |                                         ^
      131 |     const response = await request(app.getHttpServer())
      132 |       .post("/airdrop/upload/rewards")
      133 |       .set({ Authorization: `Bearer ${userJwt}` });

      at Object.<anonymous>.module.exports [as sign] (../node_modules/jsonwebtoken/sign.js:107:20)
      at JwtService.sign (../node_modules/@nestjs/jwt/dist/jwt.service.js:28:20)
      at Object.<anonymous> (airdrop.e2e-spec.ts:130:41)
```

amateima (approved):

PR #84, opened 2022-09-26
feat: add wallet and community rewards

amateima:
Changes:

- Add [models](https://github.com/across-protocol/scraper-api/tree/amatei/acx-290-add-endpoint-to-serve-wallet-rewards/src/modules/airdrop/model) for the Community Rewards and Wallet Rewards
- Add [endpoint](https://github.com/across-protocol/scraper-api/pull/84/files#diff-f91b21c92f2c60e94b5de8e3dcf5dc2b4bf31af44d40dc7c6d568ba1835f191eR13-R18) to get rewards for a particular wallet address
- Add [endpoint](https://github.com/across-protocol/scraper-api/pull/84/files#diff-f91b21c92f2c60e94b5de8e3dcf5dc2b4bf31af44d40dc7c6d568ba1835f191eR20-R42) to upload JSON files for populating the `community_rewards` and `wallet_rewards` tables
- Add a new [NestJS Guard](https://github.com/across-protocol/scraper-api/pull/84/files#diff-9acd83629deda2850757a03d3f6345dcc7e9583259d5b2ebfc795c99d941e461R13-R26) for RBAC authorization. Some endpoints (as uploading JSON rewards files) can be called only if the attached JWT has `admin` role
- Add [e2e test](https://github.com/across-protocol/scraper-api/blob/4a50e3d29d59257699a4b2f0c3d3526cb60cc288/test/airdrop.e2e-spec.ts)

### DIFF
--- a/src/modules/airdrop/module.ts
+++ b/src/modules/airdrop/module.ts
@@ -1,0 +1,21 @@
+import { Module } from "@nestjs/common";
+import { TypeOrmModule } from "@nestjs/typeorm";
+
+import { UserModule } from "../user/module";
+
+import { WalletRewardsFixture } from "./adapter/db/wallet-rewards-fixture";
+import { CommunityRewardsFixture } from "./adapter/db/community-rewards-fixture";
+import { AirdropService } from "./services/airdrop-service";
+import { AirdropController } from "./entry-points/http/controller";
+
+import { CommunityRewards } from "./model/community-rewards.entity";
+import { WalletRewards } from "./model/wallet-rewards.entity";
+import { Deposit } from "../scraper/model/deposit.entity";
+
+@Module({
+  providers: [AirdropService, WalletRewardsFixture, CommunityRewardsFixture],
+  controllers: [AirdropController],
+  imports: [TypeOrmModule.forFeature([CommunityRewards, WalletRewards, Deposit]), UserModule],
+  exports: [],
+})
+export class AirdropModule {}
--- a/src/modules/airdrop/services/airdrop-service.ts
+++ b/src/modules/airdrop/services/airdrop-service.ts
@@ -1,0 +1,196 @@
+import { Injectable, Logger } from "@nestjs/common";
+import { InjectRepository } from "@nestjs/typeorm";
+import { ethers } from "ethers";
+import { readFile } from "fs/promises";
+import BigNumber from "bignumber.js";
+import { IsNull, Not, Repository } from "typeorm";
+
+import { UserService } from "../../user/services/user.service";
+import { UserWalletService } from "../../user/services/user-wallet.service";
+
+import { Deposit } from "../../scraper/model/deposit.entity";
+import { CommunityRewards } from "../model/community-rewards.entity";
+import { WalletRewards } from "../model/wallet-rewards.entity";
+
+import { ProcessCommunityRewardsFileException, ProcessWalletRewardsFileException } from "./exceptions";
+
+@Injectable()
+export class AirdropService {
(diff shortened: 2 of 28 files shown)


src/modules/airdrop/module.ts
```diff
@@ -0,0 +1,21 @@
...
+import { CommunityRewardsFixture } from "./adapter/db/community-rewards-fixture";
+import { AirdropService } from "./services/airdrop-service";
+import { AirdropController } from "./entry-points/http/controller";
+
+import { CommunityRewards } from "./model/community-rewards.entity";
+import { WalletRewards } from "./model/wallet-rewards.entity";
+import { Deposit } from "../scraper/model/deposit.entity";
+
+@Module({
+  providers: [AirdropService, WalletRewardsFixture, CommunityRewardsFixture],
```
dohaki, line 16: The fixtures `WalletRewardsFixture` and `CommunityRewardsFixture` are only used for the e2e tests, right? I think it would make sense to only load them when running the tests in the future. Wdyt?

src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -0,0 +1,196 @@
...
+    @InjectRepository(WalletRewards) private walletRewardsRepository: Repository<WalletRewards>,
+    @InjectRepository(Deposit) private depositRepository: Repository<Deposit>,
+    private userWalletService: UserWalletService,
+    private userService: UserService,
+  ) {}
+
+  public async getRewards(walletAddress: string) {
+    const checksumAddress = ethers.utils.getAddress(walletAddress);
+    const walletRewards = await this.getWalletRewards(checksumAddress);
+    const communityRewards: string = await this.getCommunityRewards(checksumAddress);
```
dohaki, line 32: NIT: can be wrapped in a Promis.all

src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -0,0 +1,196 @@
...
+
+      const communityRewardsContent = await readFile(communityRewardsFile.path, { encoding: "utf8" });
+      const communityRewards = JSON.parse(communityRewardsContent);
+
+      for (const communityReward of communityRewards) {
+        if (communityReward["ID"] && communityReward["Total Tokens"]) {
+          await this.insertCommunityRewards(communityReward["ID"], communityReward["Total Tokens"]);
+        }
+      }
+      await this.communityRewardsRepository.delete({ processed: false });
```
dohaki, line 126: why do we need to delete here?

dohaki (approved):
Looks good 👍 Only left a few questions

src/modules/airdrop/services/airdrop-service.ts
amateima: the JSON files contains the rewards, but there is no way to know in advance if a particular address/discord id was removed comparing to the previous rewards file. So before inserting/updating the rewards I set the existing rows to be `processed = false` and on insert/update they become `processed = true`. If any rows are left with `processed = false`, it means that they are not eligible for rewards anymore and they should be deleted

src/modules/airdrop/module.ts
dohaki: Yea exactly. I guess that overlaps with the mode splitting work I am currently looking into.

amateima: Yes. It is a good idea 👍

amateima: we can have `public-api`, `scraper` and `test` as running modes, execution modes or whatever name it'll have 😄

PR #87, opened 2022-09-28
feat: split modes dynamic modules

dohaki:
Closes ACX-246

- [x] split ENV vars in .env.sample by the running modes. [See](https://github.com/across-protocol/scraper-api/pull/87/files#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049c)
- [x] edit all modules to configure the injected providers depending on the running mode

### DIFF
--- a/src/modules/scraper/service/ScraperQueuesService.ts
+++ b/src/modules/scraper/service/ScraperQueuesService.ts
@@ -59,7 +59,10 @@ export class ScraperQueuesService {
     setInterval(() => {
       for (const queueName of Object.keys(this.queuesMap)) {
         const queue = this.queuesMap[queueName] as Queue;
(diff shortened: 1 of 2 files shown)


src/app.module.ts
```diff
@@ -18,37 +18,50 @@ import { DepositModule } from "./modules/deposit/module";
...
+
+    if ([RunMode.Scraper, RunMode.Test].includes(moduleOptions.runMode)) {
+      imports.push(
+        ScraperModule.forRoot(moduleOptions),
+        BullModule.forRootAsync({
+          imports: [MessagingModule],
+          useExisting: BullConfigService,
+        }),
+      );
+    }
```
dohaki, line 57: Are these the only modules that should not be included if running in `normal` mode?

amateima (dismissed):

PR #88, opened 2022-09-29
fix: fetch community rewards by JWT claims

amateima:
`/airdrop/rewards?address=0x...`: Now, in order to fetch the community rewards component, the JWT token needs to be attached to the req

### DIFF
--- a/src/modules/airdrop/services/airdrop-service.ts
+++ b/src/modules/airdrop/services/airdrop-service.ts
@@ -26,10 +26,10 @@ export class AirdropService {
     private userService: UserService,
   ) {}
 
-  public async getRewards(walletAddress: string) {
+  public async getRewards(walletAddress: string, userId?: number) {
     const checksumAddress = ethers.utils.getAddress(walletAddress);
     const walletRewards = await this.getWalletRewards(checksumAddress);
-    const communityRewards: string = await this.getCommunityRewards(checksumAddress);
+    const communityRewards: string = await this.getCommunityRewards(userId);
     const welcomeTravellerEligible = !!walletRewards && new BigNumber(walletRewards.welcomeTravellerRewards).gt(0);
     let welcomeTravellerCompleted = false;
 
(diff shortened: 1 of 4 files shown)


src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -26,10 +26,10 @@ export class AirdropService {
     private userService: UserService,
   ) {}
 
-  public async getRewards(walletAddress: string) {
+  public async getRewards(walletAddress: string, userId?: number) {
     const checksumAddress = ethers.utils.getAddress(walletAddress);
     const walletRewards = await this.getWalletRewards(checksumAddress);
-    const communityRewards: string = await this.getCommunityRewards(checksumAddress);
+    const communityRewards: string = await this.getCommunityRewards(userId);
```
dohaki, line 32: NIT: Do we have to cast this to `string` even though it can be `undefined`? I mean are handling the `undefined` case correctly below AFAICT.

src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -177,14 +177,13 @@ export class AirdropService {
     return this.walletRewardsRepository.findOne({ where: { walletAddress } });
   }
 
-  private async getCommunityRewards(walletAddress: string): Promise<string | undefined> {
-    const userWallet = await this.userWalletService.getUserWalletByAttributes({ walletAddress });
+  private async getCommunityRewards(userId?: number): Promise<string | undefined> {
+    if (!userId) return undefined;
```
dohaki, line 181: NIT: Hmm, this looks weird to me to return `undefined` if the argument is `undefined`. Wouldn't it be cleaner to make `userId` mandatory and call this method conditionally?

dohaki (approved):
Nice, awesome that you found the issue so fast. LGTM. Left two nits though.

src/modules/airdrop/services/airdrop-service.ts
amateima: Removed the cast. We don't need it

src/modules/airdrop/services/airdrop-service.ts
amateima: 👍

PR #89, opened 2022-09-29
feat: allow editing wallet rewards

amateima:
Allow to edit the wallet rewards for a particular address. This will be enabled only on the staging environment


nicholaspai (approved):
Nice LGTM. Two follow ups
- I will make sure @Tulun and @james-a-morris see this so they can make a temp button
- how will we remove this in prod. Do we simply revert pr?

amateima:
@nicholaspai 
> * I will make sure @Tulun and @james-a-morris see this so they can make a temp button
This can be easily called from the Swagger interface

> * how will we remove this in prod. Do we simply revert pr?
This feature is hidden behind an ENV var

PR #91, opened 2022-09-30
fix: handle Discord auth error

amateima:
https://linear.app/uma/issue/ACX-327/fix-discord-authentication

The discord auth endpoint throws an error if the user is not part of the Across Discord server. 
The correct flow is to ignore silently this error and use the global user details (name and avatar)


dohaki (approved):
lgtm 👍

PR #93, opened 2022-10-18
feat: inject and serve merkle distributor claims

amateima:
https://linear.app/uma/issue/ACX-352/scraper-api-endpoint-for-handling-created-tree

Changes in this PR:
- add [MerkleDistributorRecipient](https://github.com/across-protocol/scraper-api/blob/758d5a9d732533a25cc17e3f745e95e18bcb5d8d/src/modules/airdrop/model/merkle-distributor-recipient.entity.ts) and [MerkleDistributorWindow](https://github.com/across-protocol/scraper-api/blob/758d5a9d732533a25cc17e3f745e95e18bcb5d8d/src/modules/airdrop/model/merkle-distributor-window.entity.ts) entities
- [endpoint](https://github.com/across-protocol/scraper-api/blob/758d5a9d732533a25cc17e3f745e95e18bcb5d8d/src/modules/airdrop/entry-points/http/controller.ts#L65-L77) to upload JSON file for MerkleDistributor recipients (this endpoint is intended to be called by the [merkle-distributor publish-tree script](https://github.com/across-protocol/merkle-distributor/blob/main/scripts/publish-tree.ts))
- [endpoint](https://github.com/across-protocol/scraper-api/blob/758d5a9d732533a25cc17e3f745e95e18bcb5d8d/src/modules/airdrop/entry-points/http/controller.ts#L79-L83) to serve the claims for the MerkleDistributor contract
- add [e2e tests](https://github.com/across-protocol/scraper-api/blob/758d5a9d732533a25cc17e3f745e95e18bcb5d8d/test/airdrop/merkle-distributor.e2e-spec.ts)

### DIFF
--- a/src/modules/airdrop/adapter/db/merkle-distributor-recipient.ts
+++ b/src/modules/airdrop/adapter/db/merkle-distributor-recipient.ts
@@ -1,0 +1,44 @@
+import { Injectable } from "@nestjs/common";
+import { InjectRepository } from "@nestjs/typeorm";
+import { Repository } from "typeorm";
+import { MerkleDistributorRecipient } from "../../model/merkle-distributor-recipient.entity";
+
+@Injectable()
+export class MerkleDistributorRecipientFixture {
+  public constructor(
+    @InjectRepository(MerkleDistributorRecipient)
+    private merkleDistributorRecipientRepository: Repository<MerkleDistributorRecipient>,
+  ) {}
+
+  public insertMerkleDistributorRecipient(args: Partial<MerkleDistributorRecipient> = {}) {
+    const deposit = this.merkleDistributorRecipientRepository.create(this.mockMerkleDistributorRecipientEntity(args));
+    return this.merkleDistributorRecipientRepository.save(deposit);
+  }
+
+  public insertManyMerkleDistributorRecipients(args: Partial<MerkleDistributorRecipient>[] = [{}]) {
+    const createdDeposits = this.merkleDistributorRecipientRepository.create(
+      args.map((arg) => this.mockMerkleDistributorRecipientEntity(arg)),
+    );
+    return this.merkleDistributorRecipientRepository.save(createdDeposits);
+  }
+
+  public mockMerkleDistributorRecipientEntity(
+    overrides: Partial<MerkleDistributorRecipient> = {},
+  ): Partial<MerkleDistributorRecipient> {
+    return {
+      merkleDistributorWindowId: 0,
+      address: "0x",
+      amount: "10",
+      accountIndex: 0,
+      proof: ["ox"],
+      payload: {},
+      ...overrides,
+    };
+  }
+
+  public deleteAllMerkleDistributorRecipients() {
+    return this.merkleDistributorRecipientRepository.query(
+      `truncate table "merkle_distributor_recipient" restart identity cascade`,
+    );
+  }
+}
--- a/src/modules/airdrop/entry-points/http/controller.ts
+++ b/src/modules/airdrop/entry-points/http/controller.ts
@@ -84,6 +84,7 @@ export class AirdropController {
   @Get("merkle-distributor-proof")
   @ApiTags("airdrop")
   getMerkleDistributorProof(@Query() query: GetMerkleDistributorProofQuery) {
-    return this.airdropService.getMerkleDistributorProof(query.address, query.windowIndex);
+    const includeDiscord = query.includeDiscord === "true";
+    return this.airdropService.getMerkleDistributorProof(query.address, query.windowIndex, includeDiscord);
   }
 }
--- a/src/modules/airdrop/entry-points/http/controller.ts
+++ b/src/modules/airdrop/entry-points/http/controller.ts
@@ -17,7 +17,12 @@ import { OptionalJwtAuthGuard } from "../../../auth/entry-points/http/optional-j
 import { JwtAuthGuard } from "../../../auth/entry-points/http/jwt.guard";
 import { Role, Roles, RolesGuard } from "../../../auth/entry-points/http/roles";
 import { AirdropService } from "../../services/airdrop-service";
-import { EditWalletRewardsBody, GetAirdropRewardsQuery, GetAirdropRewardsResponse, GetMerkleDistributorProofQuery } from "./dto";
+import {
+  EditWalletRewardsBody,
+  GetAirdropRewardsQuery,
+  GetAirdropRewardsResponse,
+  GetMerkleDistributorProofQuery,
+} from "./dto";
 
 @Controller("airdrop")
 export class AirdropController {
--- a/src/modules/airdrop/entry-points/http/controller.ts
+++ b/src/modules/airdrop/entry-points/http/controller.ts
@@ -6,17 +6,18 @@ import {
   Post,
   Query,
   Req,
+  UploadedFile,
   UploadedFiles,
   UseGuards,
   UseInterceptors,
 } from "@nestjs/common";
-import { FileFieldsInterceptor } from "@nestjs/platform-express";
+import { FileFieldsInterceptor, FileInterceptor } from "@nestjs/platform-express";
 import { ApiResponse, ApiTags } from "@nestjs/swagger";
 import { OptionalJwtAuthGuard } from "../../../auth/entry-points/http/optional-jwt.";
 import { JwtAuthGuard } from "../../../auth/entry-points/http/jwt.guard";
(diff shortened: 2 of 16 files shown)


src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -123,6 +132,89 @@ export class AirdropService {
...
+            .values({
+              merkleDistributorWindowId: windowId,
+              address: recipient["account"],
+              amount: recipient["amount"],
+              accountIndex: recipient["accountIndex"],
+              proof: recipient["proof"],
+              payload: recipient["metadata"],
+            })
+            .execute();
+        }
```
dohaki, line 171: AFAICT the current behavior for an edge case where an insert fails for a single recipient would require us to delete the entries manually before re-uploading the fixed file right?

dohaki: Could be nice to make this atomic somehow but maybe also over-engineered

src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -123,6 +132,89 @@ export class AirdropService {
...
+
+    return {
+      accountIndex: recipient.accountIndex,
+      address: recipient.address,
+      amount: recipient.amount,
+      payload: recipient.payload,
+      proof: recipient.proof,
+      merkleRoot: recipient.merkleDistributorWindow.merkleRoot,
+      windowIndex: recipient.merkleDistributorWindow.windowIndex,
+    };
```
dohaki, line 215: Can we also join the associated discord details (name, avatar) here from the user entity if it exists for the given address?

dohaki (commented):
Nice work 💪 I left some questions and suggestions. Let me know what you think

src/modules/airdrop/entry-points/http/controller.ts
```diff
@@ -59,4 +66,24 @@ export class AirdropController {
...
+  @ApiTags("airdrop")
+  uploadMerkleDistributorRecipients(@UploadedFile() file: Express.Multer.File) {
+    return this.airdropService.processMerkleDistributorRecipientsFile(file);
+  }
+
+  @Get("merkle-distributor-proof")
+  @ApiTags("airdrop")
+  getMerkleDistributorProof(@Query() query: GetMerkleDistributorProofQuery) {
+    return this.airdropService.getMerkleDistributorProof(query.address, query.windowIndex);
+  }
```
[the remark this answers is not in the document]
dohaki: Ok, nevermind my comment above. We won't actually handler multiple windows in the FE and assume there is only one window that should be considered.

src/modules/airdrop/model/merkle-distributor-recipient.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+  amount: string;
+
+  @Column()
+  accountIndex: number;
+
+  @Column({ type: "jsonb" })
+  proof: string[];
+
+  @Column({ type: "jsonb" })
+  payload: MerkleDistributorRecipientPayload;
```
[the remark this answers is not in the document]
amateima: it just sounded better for me. Payload sounds like "attached data", but metadata is more like "descriptive information about a piece of data (file metadata, audio metadata, etc".

src/modules/airdrop/model/merkle-distributor-window.entity.ts
```diff
@@ -0,0 +1,31 @@
...
+  rewardsToDeposit: string;
+
+  @Column()
+  merkleRoot: string;
+
+  @OneToMany(() => MerkleDistributorRecipient, (recipient) => recipient.merkleDistributorWindow)
+  recipients: MerkleDistributorRecipient;
+
+  @CreateDateColumn()
+  createdAt: Date;
```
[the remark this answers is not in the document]
amateima: This is a very good suggestion! I missed it during the implementation. I'm adding it now

src/modules/airdrop/services/airdrop-service.ts
amateima: @dohaki 
It is an atomic operation 👍 These insertions are done using a [SQL transaction](https://www.postgresql.org/docs/current/tutorial-transactions.html) (see [line 142](https://github.com/across-protocol/scraper-api/blob/d3002023cf15c479e87070430448cfa800ea984c/src/modules/airdrop/services/airdrop-service.ts#L142) where the transaction is started). If an error is thrown inside this big callback, then all the inserts are reverted. 

PS. This is the easiest way to run SQL transactions using TypeORM and basically the library handles all the operations: start, commit and revert. TypeORM can also let you handle transactions manually, [see example here](https://github.com/typeorm/typeorm/blob/master/docs/transactions.md#using-queryrunner-to-create-and-control-state-of-single-database-connection)

src/modules/airdrop/services/airdrop-service.ts
amateima: Are they needed on the FE? Can you show me a screenshot?

src/modules/airdrop/services/airdrop-service.ts
dohaki: awesome thanks for the clarification

src/modules/airdrop/services/airdrop-service.ts
dohaki: Yes sure. It is required when connecting an eligible wallet

image

src/modules/airdrop/model/merkle-distributor-window.entity.ts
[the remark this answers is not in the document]
amateima: added https://github.com/across-protocol/scraper-api/commit/e7e36ef67c75ad4b159873b8068d51fe455ee8fa

src/modules/airdrop/services/airdrop-service.ts
amateima: I see. Ok, I'll add them. It is also the option to include this data in the metadata object part of the JSON file

src/modules/airdrop/adapter/db/merkle-distributor-recipient.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  public mockMerkleDistributorRecipientEntity(
+    overrides: Partial<MerkleDistributorRecipient> = {},
+  ): Partial<MerkleDistributorRecipient> {
+    return {
+      merkleDistributorWindowId: 0,
+      address: "0x",
+      amount: "10",
+      accountIndex: 0,
+      proof: ["ox"],
```
mrice32, line 33:
```suggestion
      proof: ["0x"],
```

src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -123,6 +132,91 @@ export class AirdropService {
...
+      if (
+        error instanceof QueryFailedError &&
+        (error as QueryFailedError).driverError.constraint === "UK_merkle_distributor_window_windowIndex"
+      ) {
+        throw new DuplicatedMerkleDistributorWindowException();
+      }
+
+      throw error;
+    }
+  }
```
mrice32, line 194: I'm assuming this error just ends up causing the query to fail. Is that right?

Outside of the query failing, do we have any way of logging or reporting errors, especially bugs/internal issues (as opposed to user input issues)?

mrice32 (approved):
LGTM!

src/modules/airdrop/services/airdrop-service.ts
dohaki: Yea either that or we can just add a separate endpoint that returns the connected discord account for a given address.

dohaki: Not sure if the discord details should be part of the JSON but could also make sense yea

dohaki: I will approve for now and this can be tackled separately

src/modules/airdrop/services/airdrop-service.ts
amateima: @mrice32 
Any error caused by a failed query will make the transaction to revert. The reason I want to handle this error separately is the fact that any thrown NodeJS error will trigger a 500 Internal Server HTTP error, but for duplicated uploaded MerkleDistributor windows I want the backend to respond with a custom error (custom message and custom HTTP code). Custom HTTP errors can be created by instantiating custom HTTPException errors

> do we have any way of logging or reporting errors
For now we only log errors inside the container. Any thrown error can be seen there

PR #100, opened 2022-10-27
feat: add filled date to deposits

amateima:
This PR implements the filled date for deposits which is the timestamp of the last Fill event.

PR #101, opened 2022-10-28
feat: add statistics endpoint (stage)

amateima:
This PR add a `GET /deposits/stats` endpoint which serves the data to be displayed on the token home page

```js
{
  totalDeposits, // the total number of deposits
  avgFillTime, // the average fill time. implemented
  totalVolumeUsd, 1000000 // the total usd bridged volume
}
```

The deposits fill time was implemented previously [here](https://github.com/across-protocol/scraper-api/pull/100)

Screenshot 2022-10-28 at 11 26 36


src/modules/scraper/service.ts
```diff
@@ -1,4 +1,4 @@
-import { Injectable, Logger } from "@nestjs/common";
+import { CACHE_MANAGER, Inject, Injectable, Logger } from "@nestjs/common";
```
dohaki, line 1: nit: lint warns that these are not used. can we remove them?

dohaki (approved):
Looks good 👍 Left one nit

src/modules/scraper/service.ts
amateima: Yep. Removed 👍

PR #111, opened 2022-11-02
feat: referral view updates (stage)

dohaki:
Based on #110 as the `Claim` entities are needed, waiting until merging. Can be reviewed already though.

This PR adds the referral rewards reset functionality by updating the `deposit_referral_stats` view. The main logic addition is the following [condition to the query](https://github.com/across-protocol/scraper-api/compare/index-claims...view-updates?expand=1#diff-76c276375bb5aa6c7b38880fb43207dc2240004d7a7dc4cb9f1619a10d7aaa10R18-R23)
```sql
  AND d2."depositDate" >= (SELECT  coalesce(max(c."claimedAt"), '2000-01-01 00:00:00')
                           FROM    claim c
                           WHERE   c.account = d2."stickyReferralAddress"
                           AND     c."windowIndex" >= ${
                             configValues().web3.merkleDistributor.referralsStartWindowIndex
                           })
```

I also ported the view queries into [TypeORM's `ViewEntity` decorator](https://typeorm.io/view-entities) so that migration files can be better generated. Also because I somehow needed to pass in a config value into the query. In the future, we can use these models for some of our referral queries.


src/modules/referral/services/queries.ts
```diff
@@ -122,3 +122,7 @@ export const updateStickyReferralAddressesForDepositor = () => {
     where deposit.id = d4.id;
   `;
 };
+
+export const getRefreshMaterializedView = () => {
+  return `REFRESH MATERIALIZED VIEW deposits_mv`;
+};
```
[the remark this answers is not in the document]
dohaki: So there are "normal" views and "materialized" views:
- Normal views are populated each time you access / query them. We use such a view [here](https://github.com/across-protocol/scraper-api/pull/111/files#diff-76c276375bb5aa6c7b38880fb43207dc2240004d7a7dc4cb9f1619a10d7aaa10R5-R33). Only the query/definition is stored in the DB but the data not.
- Materialized views are kinda cached views. The initial data get populated when you create it but needs to be refreshed (either manually or automatically). The cached data is I believe stored in the DB but might be out of sync (depending on when it was last refreshed).

nicholaspai (approved):

nicholaspai (approved):

migrations/1667412728068-DepositReferralStats.ts
```diff
@@ -0,0 +1,114 @@
...
+        "VIEW",
+        "deposit_referral_stats",
+        'SELECT d1.id,\n      count(distinct d2."depositorAddr") as "referralCount",\n      sum((d2.amount / power(10, t.decimals)) * hmp.usd) as "referralVolume"\n    FROM deposit d1\n    LEFT JOIN deposit d2 on d1."stickyReferralAddress" = d2."stickyReferralAddress"\n      AND d1."depositDate" >= d2."depositDate"\n      AND d1."stickyReferralAddress" is not null\n      AND d1."depositDate" is not null\n      AND d1."tokenId" is not null\n      AND d1."priceId" is not null\n      AND d1.status = \'filled\'\n      AND d2."depositDate" >= (SELECT  coalesce(max(c."claimedAt"), \'2000-01-01 00:00:00\')\n                               FROM    claim c\n                               WHERE   c.account = d2."stickyReferralAddress"\n                               AND     c."windowIndex" >= 1)\n    JOIN historic_market_price hmp on d2."priceId" = hmp.id\n    JOIN token t on d2."tokenId" = t.id\n    WHERE d1."stickyReferralAddress" is not null\n      AND d1."depositDate" is not null\n      AND d1."tokenId" is not null\n      AND d1."priceId" is not null\n      AND d1.status = \'filled\'\n    GROUP BY d1.id',
+      ],
+    );
```
amateima, line 50: Not sure what this query does

migrations/1667412728068-DepositReferralStats.ts
```diff
@@ -0,0 +1,114 @@
...
+      AND d1."depositDate" >= d2."depositDate"
+      AND d1."stickyReferralAddress" is not null
+      AND d1."depositDate" is not null
+      AND d1."tokenId" is not null
+      AND d1."priceId" is not null
+      AND d1.status = 'filled'
+      AND d2."depositDate" >= (SELECT  coalesce(max(c."claimedAt"), '2000-01-01 00:00:00')
+                               FROM    claim c
+                               WHERE   c.account = d2."stickyReferralAddress"
+                               AND     c."windowIndex" >= 1)
```
amateima, line 32: Hmm, I might be wrong, but this should not be the time of the claimed event, but the timestamp when the referral rewards were computed for the airdrop, right?

amateima (changes requested):
I have one question about the updated definition of the VIEW

migrations/1667412728068-DepositReferralStats.ts
dohaki: Yes, you are right. So I need to link the `MerkleDistributorWindow` entity here, I assume. But which timestamp do we consider? 🤔 I think we need to add additional payload to the tree metadata for referral rewards, something like `snapshotAt` and use this.

migrations/1667412728068-DepositReferralStats.ts
dohaki: It was auto-generated by TypeORM

migrations/1667412728068-DepositReferralStats.ts
dohaki: When we do these referral reward snapshots, we need to make sure to somehow only include deposits with a high number of confirmations to prevent double rewards for re-orged deposits.

amateima (approved):

PR #139, opened 2022-11-21
feat: add V1 volume and deposits count to /deposits/stats (stage)

amateima:


src/modules/deposit/service.ts
```diff
@@ -6,7 +6,7 @@ import { Cache } from "cache-manager";
 import { Deposit } from "../scraper/model/deposit.entity";
 import { getAvgFillTimeQuery, getTotalDepositsQuery, getTotalVolumeQuery } from "./adapter/db/queries";
 
-export const GENERAL_STATS_CACHE_KEY = "stats:general";
+export const DEPOSITS_STATS_CACHE_KEY = "deposits:stats";
```
amateima, line 9: Found a better naming for the cache key

nicholaspai (approved):
I assume these are numbers that Chase sent you so LGTM

nicholaspai:
Also @amateima it looks like there are some linting options
Screen Shot 2022-11-21 at 06 03 51

amateima:
> I assume these are numbers that Chase sent you so LGTM

Yes

PR #140, opened 2022-11-21
feat: filter out pending deposits older than 1 day

amateima:
https://linear.app/uma/issue/ACX-451/all-txns-page-hide-stuck-pending-deposits

This PR filters out pending deposits older than 1 day

### DIFF
--- a/src/modules/deposit/service.ts
+++ b/src/modules/deposit/service.ts
@@ -1,7 +1,7 @@
 import { DateTime } from "luxon";
 import { CACHE_MANAGER, Inject, Injectable } from "@nestjs/common";
 import { InjectRepository } from "@nestjs/typeorm";
-import { Repository, IsNull, Not } from "typeorm";
+import { Repository } from "typeorm";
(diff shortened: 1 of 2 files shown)


src/modules/deposit/service.ts
```diff
@@ -35,18 +35,30 @@ export class DepositService {
...
+      [deposits, total] = await this.depositRepository
+        .createQueryBuilder("d")
+        .where("d.status = :status", { status })
+        .andWhere("d.depositDate is not null")
+        .orderBy("d.depositDate", "DESC")
+        .take(limit)
+        .skip(offset)
+        .getManyAndCount();
+    } else if (status === "pending") {
+      // filter out pending deposits older than 1 day
```
nicholaspai, line 52:
```suggestion
      // filter out pending deposits older than 1 day because the relayer will ignore such deposits using its fixed lookback
```

nicholaspai (approved):
One suggestion ! LGTM

PR #143, opened 2022-11-21
fix: revert bt eligible (prod)

amateima:
This PR reverts the query which returns the addresses eligible for BT rewards and also applies an upper bound date to eligible BT deposits


cc7768 (approved):
This looks like the correct reversion to me

Do we want to have an upper bound on the dates though? I think we made the original change prior to cutting off the BT.

PR #144, opened 2022-11-21
fix: apply upper bound to BT eligible deposits (stage)

amateima:


src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -104,14 +105,15 @@ export class AirdropService {
     let welcomeTravellerCompleted = false;
 
     if (welcomeTravellerEligible) {
-      const depositCountQuery = await this.depositRepository
+      const depositCountQuery = this.depositRepository
```
nicholaspai, line 108: was this `await` a mistake?

nicholaspai (approved):

src/modules/airdrop/services/airdrop-service.ts
amateima: not a mistake, but it was not necessary.

PR #151, opened 2022-11-23
fix: add test for BT rewards

amateima:
This PR adds a test to ensure that deposits after a certain date doesn't count for BT rewards `completed` flag


dohaki (approved):
Looks good

PR #158, opened 2022-11-28
feat: add memory caching to proofs endpoints (stage)

amateima:


src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -268,10 +279,17 @@ export class AirdropService {
           }
         : null,
     };
+
+    await this.cacheManager.set(cacheKey, data, 300);
```
amateima, line 283: cache the result before returning the response

amateima, line 283: cache the result before returning the response

src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -293,6 +311,9 @@ export class AirdropService {
       ipfsHash: recipient.merkleDistributorWindow.ipfsHash || null,
       discord: null,
     }));
+
+    await this.cacheManager.set(cacheKey, data, 300);
```
amateima, line 315: cache the result before returning the response

dohaki (approved):

PR #159, opened 2022-11-28
feat: add memory caching to proofs endpoint (prod)

amateima:
This PR caches for 5 minutes the response of the endpoint that returns the proofs for the merkle distributor contract


nicholaspai (approved):

dohaki (approved):

src/modules/airdrop/services/airdrop-service.ts
```diff
@@ -268,10 +279,17 @@ export class AirdropService {
           }
         : null,
     };
+
+    await this.cacheManager.set(cacheKey, data, 300);
```
nicholaspai, line 283: I'd recommend putting the two 300 hardcoded values into an easier to change variable

amateima: yeah, you're right

PR #164, opened 2022-11-29
fix: clear HTTP logs

amateima:
For each req the backend logs a message similar to `${method} ${endpoint} ${statusCode} ${statusMessage}`

This PR allows only 5xx and 4xx HTTP logs. Displaying logs on each requests make them unreadable during periods with high traffic

### DIFF
--- a/src/logging.interceptor.ts
+++ b/src/logging.interceptor.ts
@@ -19,8 +19,6 @@ export class LogsMiddleware implements NestMiddleware {
       if (statusCode >= 400) {
         return this.logger.warn(message);
       }
-
-      return this.logger.log(message);
     });
 
     next();


src/logging.interceptor.ts
```diff
@@ -19,8 +19,6 @@ export class LogsMiddleware implements NestMiddleware {
       if (statusCode >= 400) {
         return this.logger.warn(message);
       }
-
-      return this.logger.log(message);
```
dohaki, line 23: NIT: Ideally, this would be configurable by setting a `LOG_LEVEL` env var or similar. Not sure how this works in NestJS though. Something we can look into later.

dohaki (approved):
LGTM. Left one remark we can look into later.

PR #168, opened 2022-12-06
fix: calculate average fill time only for the last 24 hours

amateima:
Decrease the time window for calculating the average fill time from 31 days to 24 hours


dohaki (approved):
LGTM !

PR #173, opened 2022-12-09
feat: 💲 integrate ACX price into referral rewards calculation

amateima:
This PR updates the calculation of the referral rewards to use the `acxUsdPrice` column of the `deposit` table, instead of the hardcoded 0.1 value. [Here](https://github.com/across-protocol/scraper-api/pull/173/files#diff-435286f80a01293ba2937508138d0b82e9083a53499b7f2caafff986d2de05c8) you can see the updated queries

### DIFF
--- a/migrations/1669984610620-DepositsMv.ts
+++ b/migrations/1669984610620-DepositsMv.ts
@@ -1,0 +1,81 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class DepositsMv1669984610620 implements MigrationInterface {
+  name = "DepositsMv1669984610620";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP MATERIALIZED VIEW IF EXISTS "deposits_mv"`);
+    await queryRunner.query(`DROP VIEW IF EXISTS "deposits_filtered_referrals"`);
+    await queryRunner.query(`CREATE VIEW "deposits_filtered_referrals" AS 
+        SELECT 
+        d.id,
+        d."depositorAddr",
+        d."stickyReferralAddress",
+        d."depositDate",
+        d."priceId",
+        d."tokenId",
+        d.amount,
+        d."rewardsWindowIndex",
+        case when d."rewardsWindowIndex" = c."windowIndex" then d."rewardsWindowIndex" else -1 end as "referralClaimedWindowIndex",
+        hmp."usd",
+        t.decimals
+      FROM deposit d
+      JOIN historic_market_price hmp on d."priceId" = hmp.id
+      JOIN token t on d."tokenId" = t.id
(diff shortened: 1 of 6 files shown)


migrations/1669984610620-DepositsMv.ts
```diff
@@ -0,0 +1,81 @@
...
+      FROM deposit d
+      JOIN historic_market_price hmp on d."priceId" = hmp.id
+      JOIN token t on d."tokenId" = t.id
+      LEFT JOIN claim c on d."rewardsWindowIndex" = c."windowIndex" and d."stickyReferralAddress" = c."account"
+      WHERE "stickyReferralAddress" is not null
+        AND "depositDate" is not null
+        AND "tokenId" is not null
+        AND "priceId" is not null
+        AND status = 'filled'
+        AND d."acxUsdPrice" is not null;
```
amateima, line 31: this is the only line that changed in the SQL `deposits_filtered_referrals` VIEW that contains the deposits eligible for referral rewards calculations

migrations/1669984610620-DepositsMv.ts
```diff
@@ -0,0 +1,81 @@
...
+        d."depositorAddr",
+        d."rewardsWindowIndex",
+        case when d."rewardsWindowIndex" = c."windowIndex" and d."depositorAddr" = c.account then d."rewardsWindowIndex" else -1 end as "depositorClaimedWindowIndex",
+        d1."referralClaimedWindowIndex",
+        d."stickyReferralAddress" AS "referralAddress",
+        d."depositDate",
+        hmp.usd AS "tokenUsdPrice",
+        (d."realizedLpFeePctCapped" / power(10, 18)) * (d.amount / power(10, t.decimals)) * hmp.usd AS "realizedLpFeeUsd",
+        (d."bridgeFeePct" / power(10, 18)) * (d.amount / power(10, t.decimals)) * hmp.usd AS "bridgeFeeUsd",
+        d."acxUsdPrice",
```
amateima, line 53: This is the only line that changed in the `deposits_mv` MATERIALIZED VIEW definition. THe `acxUsdPrice` needs to be exposed because the referral rewards amounts are computed based on it instead of using the hardcoded 0.1 value.

nicholaspai (approved):
LGTM but have one comment

src/modules/referral/services/service.ts
```diff
@@ -242,7 +242,7 @@ export class ReferralService {
           .multipliedBy(feePct)
           .multipliedBy(d.multiplier)
           .multipliedBy(new BigNumber(10).pow(18))
-          .dividedBy(this.appConfig.values.acxUsdPrice)
+          .dividedBy(d.acxUsdPrice)
```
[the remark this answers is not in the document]
amateima: No, because only the [deposits with this column not null](https://github.com/across-protocol/scraper-api/pull/173/files#diff-ffeb9268c58aa039146f3c129f010a71b2dea4e103bbda111e697352a785dc8bR31) are considered

PR #176, opened 2022-12-12
feat: suggested fees consumer

dohaki:
Closes ACX-465

This PR adds a consumer that fetches the suggested fees from the Vercel API at the time a deposit gets indexed by the Scraper. If the deposit was made in the past, we check whether the deposit date is within a configurable threshold and fallback to 1bp if not.

### DIFF
--- a/src/modules/scraper/adapter/messaging/SuggestedFeesConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/SuggestedFeesConsumer.ts
@@ -1,0 +1,73 @@
+import { OnQueueFailed, Process, Processor } from "@nestjs/bull";
+import { Logger } from "@nestjs/common";
+import { Job } from "bull";
+import { InjectRepository } from "@nestjs/typeorm";
+import { Repository } from "typeorm";
+import { DateTime } from "luxon";
+import { utils } from "ethers";
+
+import { SuggestedFeesService } from "../../../suggested-fees/service";
+import { SuggestedFeesQueueMessage, ScraperQueue } from ".";
+import { Deposit } from "../../model/deposit.entity";
+import { AppConfig } from "../../../configuration/configuration.service";
+
+@Processor(ScraperQueue.SuggestedFees)
+export class SuggestedFeesConsumer {
+  private logger = new Logger(SuggestedFeesConsumer.name);
+
+  constructor(
+    private appConfig: AppConfig,
+    private suggestedFeesService: SuggestedFeesService,
+    @InjectRepository(Deposit) private depositRepository: Repository<Deposit>,
+  ) {}
+
+  @Process()
+  private async process(job: Job<SuggestedFeesQueueMessage>) {
+    const { depositId } = job.data;
+
+    const deposit = await this.depositRepository.findOne({
+      where: { id: depositId },
+    });
+
+    if (!deposit) {
+      this.logger.verbose(`${ScraperQueue.SuggestedFees} deposit with id '${depositId}' does not exist in db`);
+      return;
+    }
+
+    if (deposit.suggestedRelayerFeePct) {
+      this.logger.verbose(
+        `${ScraperQueue.SuggestedFees} deposit with id '${depositId}' already has a suggested fees entry`,
+      );
+      return;
+    }
+
+    if (!deposit.depositDate) {
+      throw new Error(`Deposit with id '${depositId}' needs 'depositDate' entry in order to fetch suggested fees`);
+    }
+
+    let suggestedRelayerFeePct: string;
+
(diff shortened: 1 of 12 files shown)


src/modules/scraper/module.ts
```diff
@@ -41,6 +44,7 @@ import { DepositAcxPriceConsumer } from "./adapter/messaging/DepositAcxPriceCons
     TokenPriceConsumer,
     DepositFilledDateConsumer,
     DepositAcxPriceConsumer,
+    SuggestedFeesConsumer,
```
amateima, line 47: Was this added twice in the providers list?

src/modules/suggested-fees/service.ts
```diff
@@ -0,0 +1,38 @@
...
+    relayFeePct: string;
+    relayFeeTotal: string;
+    lpFeePct: string;
+    timestamp: string;
+    isAmountTooLow: boolean;
+  };
+};
+
+@Injectable()
+export class SuggestedFeesService {
```
amateima, line 28: I don't think we need a separate module for this service. Actually you thought the same since you added this service to the providers list of the `ScraperModule`. We can move this file to `modules/scraper/service/suggested-fees-service.ts` or whatever. 
I also see `adapter` subdirectory as the place for putting the code for interacting with external actors: messaging system, database, so this piece code can also be placed at `modules/scraper/adapter/across-serverless-api/[some-service-filename]`

src/modules/scraper/adapter/messaging/SuggestedFeesConsumer.ts
```diff
@@ -0,0 +1,73 @@
...
+      this.logger.verbose(`${ScraperQueue.SuggestedFees} deposit with id '${depositId}' does not exist in db`);
+      return;
+    }
+
+    if (deposit.suggestedRelayerFeePct) {
+      this.logger.verbose(
+        `${ScraperQueue.SuggestedFees} deposit with id '${depositId}' already has a suggested fees entry`,
+      );
+      return;
+    }
```
amateima, line 42: I think we should not stop the code's execution in this case. What if we change the logic for computing the suggested fees and we want to re-run this logic for all deposits, including the ones that have this field populated?

src/modules/scraper/adapter/messaging/SuggestedFeesConsumer.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    if (!deposit.depositDate) {
+      throw new Error(`Deposit with id '${depositId}' needs 'depositDate' entry in order to fetch suggested fees`);
+    }
```
amateima, line 46: 👍

src/modules/scraper/adapter/messaging/SuggestedFeesConsumer.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    let suggestedRelayerFeePct: string;
+
+    const diffToNowHours = DateTime.fromJSDate(deposit.depositDate).diffNow().as("hours");
```
amateima, line 50: I don't think I discussed with you adding the Luxon library for managing dates, so 👍  for using it. I'm just curious, where did you find the api of `diffNow` and `as`? I couldn't find it in the docs when I needed it and it is a very nice way to compute date differences.

amateima (approved):
This looks very good! Adding consumers and queues requires some steps to be made: register the queue, register the consumer, log queue stats, etc and none of those were skipped 👍 💯

src/modules/scraper/module.ts
dohaki: No, one is the service `SuggestedFeesService` and the other one the consumer `SuggestedFeesConsumer`

src/modules/scraper/adapter/messaging/SuggestedFeesConsumer.ts
dohaki: Good point. Will remove this if clause

src/modules/scraper/adapter/messaging/SuggestedFeesConsumer.ts
dohaki: I also use luxon quite a lot personally :D Yea I initially discovered them through Intellisense actually. But you can also find them in the API docs
- https://moment.github.io/luxon/api-docs/index.html#datetimediffnow
- https://moment.github.io/luxon/api-docs/index.html#durationas

src/modules/scraper/module.ts
amateima: 👍

src/modules/scraper/adapter/messaging/SuggestedFeesConsumer.ts
amateima: nicee. Thx 🙌

PR #177, opened 2022-12-12
feat: remove referral bonus

amateima:
Starting with 2022-12-13, we use the real ACX price for computing the referral rewards, so the referral bonus needs to be removed

### DIFF
--- a/migrations/1669984610621-DepositsMv.ts
+++ b/migrations/1669984610621-DepositsMv.ts
@@ -1,0 +1,55 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class DepositsMv1669984610621 implements MigrationInterface {
+  name = "DepositsMv1669984610621";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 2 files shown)


migrations/1669984610621-DepositsMv.ts
```diff
@@ -0,0 +1,55 @@
...
+        WHEN d1."referralCount" >= 10 OR d1."referralVolume" >= 250000 THEN 0.7
+        WHEN d1."referralCount" >= 5 OR d1."referralVolume" >= 100000 THEN 0.6
+        WHEN d1."referralCount" >= 3 OR d1."referralVolume" >= 50000 THEN 0.5
+        ELSE 0.4
+      END AS "referralRate",
+      CASE
+          WHEN d."depositDate" < '2022-07-22 17:00:00' THEN 3
+          WHEN d."depositDate" < '2022-12-13 00:00:00' THEN 2
+          ELSE 1
+      END AS multiplier
```
amateima, line 39: These are only lines that have changed in the `MATERIALIZED VIEW` definition

dohaki (approved):
LGTM 👍

PR #178, opened 2022-12-12
test(e2e): try to fix caching in tests

dohaki:
@amateima I was trying to fix the e2e tests but not sure why some are still failing 🤔 Any ideas?

Fixes ACX-462


amateima:
https://linear.app/uma/issue/ACX-462/fix-claim-e2e-tests

amateima:
@dohaki 
yeah…we can skip calling `cacheManager.set()` entirely and return the data directly if cache time is 0. What do you think?

dohaki:
> @dohaki yeah…we can skip calling `cacheManager.set()` entirely and return the data directly if cache time is 0. What do you think?

Yes makes sense to me. I changed it in this [commit](https://github.com/across-protocol/scraper-api/pull/178/commits/7855b8f9bd6ae28a7eaed7ffe5f0e2a443468bb8) but two test cases are still failing 🤔 The reset is not working correctly after a claim.

amateima:
> > @dohaki yeah…we can skip calling `cacheManager.set()` entirely and return the data directly if cache time is 0. What do you think?
> 
> 
> 
> Yes makes sense to me. I changed it in this [commit](https://github.com/across-protocol/scraper-api/pull/178/commits/7855b8f9bd6ae28a7eaed7ffe5f0e2a443468bb8) but two test cases are still failing 🤔 The reset is not working correctly after a claim.

I'm checking it

amateima (approved):

PR #179, opened 2022-12-13
feat: handle speed up events

dohaki:
Closes ACX-466

This adds the support of `RequestedSpeedUpDeposit` events in the Scraper API by adding the respective consumer and two new columns to the deposit entity
- intialRelayerFeePct
- speedUps

Additionally, the existing `depositRelayerFeePct` column reflects the latest relayer fee pct.


src/modules/configuration/index.ts
```diff
@@ -95,7 +95,7 @@ export const configValues = () => ({
   },
   suggestedFees: {
     apiUrl: process.env.SUGGESTED_FEES_API_URL || "https://across.to/api/suggested-fees",
-    fallbackThresholdHours: Number(process.env.SUGGESTED_FEES_FALLBACK_THRESHOLD_HOURS || "24"),
+    fallbackThresholdHours: Number(process.env.SUGGESTED_FEES_FALLBACK_THRESHOLD_HOURS || "4"),
```
dohaki, line 98: I changed the default to 4 hours because 24 seemed a bit too high.

src/modules/scraper/adapter/messaging/SpeedUpEventsConsumer.ts
```diff
@@ -0,0 +1,65 @@
...
+    const { transactionHash, newRelayerFeePct, blockNumber, depositSourceChainId } = data;
+
+    deposit.speedUps = [
+      ...deposit.speedUps,
+      { hash: transactionHash, newRelayerFeePct, blockNumber, depositSourceChainId },
+    ].sort((a, b) => b.blockNumber - a.blockNumber);
+
+    deposit.depositRelayerFeePct = deposit.speedUps[0].newRelayerFeePct;
+
+    return this.depositRepository.save(deposit);
```
amateima, line 52: Can we use `.update` instead of `.save()`? I will update `FillEventsConsumer` to use the same approach

amateima (changes requested):

src/modules/scraper/adapter/messaging/SpeedUpEventsConsumer.ts
dohaki: done

amateima (approved):

PR #182, opened 2022-12-14
feat: endpoint to get user deposits

dohaki:
Closes ACX-506

Adds a new endpoint to be used on the FE on the "My Transactions" page. More info in the second approach section of t[his Linear task](https://linear.app/uma/issue/ACX-468/handle-new-unprofitable-condition-on-my-transactions-page).

### DIFF
--- a/src/modules/deposit/controller.ts
+++ b/src/modules/deposit/controller.ts
@@ -1,4 +1,4 @@
-import { Controller, Get, Param, Query } from "@nestjs/common";
+import { Controller, Get, Query } from "@nestjs/common";
 import { ApiResponse, ApiTags } from "@nestjs/swagger";
 import { DepositService } from "./service";
 import { GetDepositsQuery, GetDepositsStatsResponse } from "./dto";
--- a/src/modules/deposit/controller.ts
+++ b/src/modules/deposit/controller.ts
@@ -1,5 +1,6 @@
-import { Controller, Get, Query } from "@nestjs/common";
+import { Controller, Get, Param, Query } from "@nestjs/common";
 import { ApiResponse, ApiTags } from "@nestjs/swagger";
+import { utils } from "ethers";
 import { DepositService } from "./service";
 import { GetDepositsQuery, GetDepositsStatsResponse } from "./dto";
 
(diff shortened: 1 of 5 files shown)


nicholaspai (commented):
Do you think this is going to significantly increase load on the scraper api from requests sent by front end?

src/modules/deposit/controller.ts
```diff
@@ -1,5 +1,6 @@
-import { Controller, Get, Query } from "@nestjs/common";
+import { Controller, Get, Param, Query } from "@nestjs/common";
 import { ApiResponse, ApiTags } from "@nestjs/swagger";
+import { utils } from "ethers";
```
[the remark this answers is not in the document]
dohaki: Removed!

dohaki:
> Do you think this is going to significantly increase load on the scraper api from requests sent by front end?

@nicholaspai I don't think that this would significantly increase the load... Definitely more because every time the user visits the "My Transactions", we will this endpoint. But AFAICT the performance of this query shouldn't become a problem. Correct me if I am wrong @amateima

src/modules/deposit/controller.ts
```diff
@@ -15,6 +15,14 @@ export class DepositController {
     return this.depositService.getDeposits(query.status, limit, offset);
   }
 
+  @Get("deposits/:address")
```
amateima, line 18: What do you think if we make the `address` a query parameter as well? This means that we can re-use the endpoint from above. Usually endpoint parameters are used to identify resources (by an unique id for example). IN our case the address is more like a filtering field

src/modules/deposit/service.ts
```diff
@@ -36,6 +38,44 @@ export class DepositService {
     return data;
   }
 
+  public async getUserDeposits(userAddress: string, status?: "filled" | "pending", limit = 10, offset = 0) {
+    let userDeposits: Deposit[] = [];
+    let total = 0;
+
+    if (!utils.isAddress(userAddress)) {
```
amateima, line 45: The addresses are stored in the DB in checksum form, so let's transform the `userAddress` by calling `ethers.utils.getAddress()` in a `try catch` and throw `InvalidAddressException` if `getAddress()` fails

src/modules/deposit/controller.ts
dohaki: Yea that is a good point. Will change this

dohaki: Changed it [here](https://github.com/across-protocol/scraper-api/pull/182/commits/0dade652216858a0130582b1c0cdd3975942c3a3)

src/modules/deposit/service.ts
dohaki: Changed it [here](https://github.com/across-protocol/scraper-api/pull/182/commits/0dade652216858a0130582b1c0cdd3975942c3a3)

amateima (approved):

PR #183, opened 2022-12-19
feat: reset referral rewards after claim

amateima:
This PR updates the logic for computing the referral rewards by grouping the deposits by the Claim events.

### DIFF
--- a/migrations/1671299400000-Deposit.ts
+++ b/migrations/1671299400000-Deposit.ts
@@ -1,0 +1,16 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Deposit1671299400000 implements MigrationInterface {
+  name = "Deposit1671299400000";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      create index "IX_deposit_srAddress_depositDate_pId_tId_status" 
+        on deposit ("stickyReferralAddress", "depositDate", "priceId", "tokenId", status)
+    `);
+  }
+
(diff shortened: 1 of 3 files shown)


migrations/1671299400000-Deposit.ts
```diff
@@ -0,0 +1,16 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Deposit1671299400000 implements MigrationInterface {
+  name = "Deposit1671299400000";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      create index "IX_deposit_srAddress_depositDate_pId_tId_status" 
+        on deposit ("stickyReferralAddress", "depositDate", "priceId", "tokenId", status)
+    `);
```
amateima, line 10: This index will help to get the deposits faster when using these fields in the WHERE clause

migrations/1671299400000-Deposit.ts
```diff
@@ -0,0 +1,16 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Deposit1671299400000 implements MigrationInterface {
```
pxrl, line 3: Just curious: what's the significance of this timestamp?

`date -u -r 1671299400 
Sat Dec 17 17:50:00 UTC 2022`

amateima: this is the timestamp when the migration file was created. The migration files contain the queries that define the structure of the database: creating tables, adding columns, altering tables, etc, so it's crucial for the migration files to run chronologically

PR #186, opened 2022-12-27
feat: track fill event amplitude

dohaki:
Closes ACX-532


src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
```diff
@@ -0,0 +1,94 @@
...
+    const amountFormatted = utils.formatUnits(deposit.amount, deposit.token.decimals);
+    const weiPctValuesFormatter = makeWeiPctValuesFormatter(amountFormatted, deposit.price.usd);
+    const formattedLpFeeValues = weiPctValuesFormatter(deposit.realizedLpFeePct);
+    const formattedRelayFeeValues = weiPctValuesFormatter(deposit.depositRelayerFeePct);
+    const formattedBridgeFeeValues = weiPctValuesFormatter(deposit.bridgeFeePct);
+
+    this.trackService.trackDepositFilledEvent(deposit.depositorAddr, {
+      capitalFeePct: "-",
+      capitalFeeTotal: "-",
+      capitalFeeTotalUsd: "-",
```
dohaki, line 44: Is it possible to derive these values at the time a fill event is indexed?

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
```diff
@@ -0,0 +1,94 @@
...
+      fromChainId: String(deposit.sourceChainId),
+      fromChainName: ChainIdToName[deposit.sourceChainId],
+      fromTokenAddress: deposit.tokenAddr,
+      isAmountTooLow: false,
+      lpFeePct: formattedLpFeeValues.pct,
+      lpFeeTotal: formattedLpFeeValues.total,
+      lpFeeTotalUsd: formattedLpFeeValues.totalUsd,
+      NetworkFeeNative: "-",
+      NetworkFeeNativeToken: "-",
+      NetworkFeeUsd: "-",
```
dohaki, line 56: Not sure which transaction to take here in the case of multiple fill transactions 🤔 Also, do I need to call `providerDestinationChain.getTransaction(txHash)` to get these values?

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
```diff
@@ -0,0 +1,94 @@
...
+      recipient: deposit.depositorAddr, // TODO: extend deposit entity to also index `recipient`
```
dohaki, line 57: We currently do not index the `recipient` address of a `FilledRelay` event. Would it a better approach to extend the deposit entity or pass this field as part of the queue message?

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
```diff
@@ -0,0 +1,94 @@
...
+      referralProgramAddress: deposit.referralAddress || "-",
+      relayFeePct: formattedRelayFeeValues.pct,
+      relayFeeTotal: formattedRelayFeeValues.total,
+      relayFeeTotalUsd: formattedRelayFeeValues.totalUsd,
+      relayGasFeePct: "-",
+      relayGasFeeTotal: "-",
+      relayGasFeeTotalUsd: "-",
```
dohaki, line 64: Basically same questions as here: https://github.com/across-protocol/scraper-api/pull/186/files#r1057668898

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
```diff
@@ -0,0 +1,94 @@
...
+        .minus(formattedBridgeFeeValues.total)
+        .multipliedBy(deposit.price.usd)
+        .toFixed(2),
+      toChainId: String(deposit.destinationChainId),
+      toChainName: ChainIdToName[deposit.destinationChainId],
+      tokenSymbol: deposit.token.symbol,
+      totalBridgeFee: formattedBridgeFeeValues.total,
+      totalBridgeFeePct: formattedBridgeFeeValues.pct,
+      totalBridgeFeeUsd: formattedBridgeFeeValues.totalUsd,
+      toTokenAddress: deposit.tokenAddr, // TODO: retrieve correct address
```
dohaki, line 83: How do I get this value on the Scraper API?

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
```diff
@@ -0,0 +1,94 @@
...
+      transactionHash: deposit.fillTxs[0].hash, // TODO: determine which tx hash to use
```
dohaki, line 84: Which tx hash should be used here?

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
amateima: I would extend the deposit entity and populate the field using the date from `FundsDeposited` event

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: @nicholaspai

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: @nicholaspai

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: Use hardcoded map as in FE

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: Use latest fill tx hash

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
nicholaspai: You cannot exactly. You can get close if you know the gas price at the time the event was indexed. The [capital fee is a math formula](https://github.com/across-protocol/sdk-v2/blob/master/src/relayFeeCalculator/relayFeeCalculator.ts#L123) that isn't time dependent but it depends on a `capitalCostsConfig` that is stored on the API that quotes the fee.  The gas fee depends on the [gas price and gas token price](https://github.com/across-protocol/sdk-v2/blob/master/src/relayFeeCalculator/relayFeeCalculator.ts#L103) at the time of indexing

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
nicholaspai: shouldn't you track every fill transaction  since there can be many? And yes I think you can call [getTransaction](https://docs.ethers.org/v5/api/providers/provider/#Provider-getTransaction)

dohaki: Yea that's what I am not sure about. I understood that this event should track the event when a deposit gets fully filled. I will clarify with Ryan on this

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: I ended up using a value from the `FilledRelay` event

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: It got clarified and as you said all fill events should be tracked with the respective fill values.

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (!fillTx) {
+      throw new Error("Fill tx does not exist on deposit");
+    }
+
+    if (!fillTx.date) {
+      throw new Error("Fill tx does not have a date");
+    }
+
+    const destinationChainInfo = chainIdToInfo[deposit.destinationChainId];
+    const sourceChainInfo = chainIdToInfo[deposit.sourceChainId];
```
amateima, line 64: I think we need to handle the case when sourceChainInfo and destinationChainInfo are `undefined`

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
```diff
@@ -0,0 +1,167 @@
...
+      toTokenAddress: utils.getAddress(destinationToken),
+      transactionHash: fillTx.hash,
+      transferCompleteTimestamp: String(DateTime.fromISO(fillTx.date).toMillis()),
+      transferQuoteBlockNumber: String(fillTxBlock.blockNumber),
+    });
+  }
+
+  private async getFillTxNetworkFee(destinationChainId: number, fillTxHash: string) {
+    const destinationChainProvider = this.providers.getProvider(destinationChainId);
+    const destinationChainInfo = chainIdToInfo[destinationChainId];
```
amateima, line 143: destinationChainInfo might be `undefined`

amateima (changes requested):

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: We now derive it by subtracting the tx fee from the relayer fee. See https://github.com/across-protocol/scraper-api/pull/186/commits/50eb98713e8ff05b07b532b00ce1971e55220b97

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: Yea good point. I will add a change that defaults to `"-"` if chain info is unknown.

dohaki: https://github.com/across-protocol/scraper-api/pull/186/commits/a310e99351d2e2e628674b4706d311c914965312

src/modules/scraper/adapter/messaging/TrackFillEventConsumer.ts
dohaki: https://github.com/across-protocol/scraper-api/pull/186/commits/a310e99351d2e2e628674b4706d311c914965312

amateima (approved):

PR #191, opened 2023-01-02
fix: fix grouping deposits by claimed window

amateima:
Consider only window > 0 when grouping deposits by claimed window for computing referral rewards


dohaki (approved):
LGTM

PR #195, opened 2023-01-06
fix: fix referral address extraction

amateima:
`subtractFunctionArgsFromCallData` function removes the function arguments from the calldata, returning only the arbitrary data that can be attached to a transaction. IF the calldata's length in bytes is less than the length of the deposit function signature `["address", "address", "uint256", "uint256", "uint64", "uint32"]` , then return empty string.

### DIFF
--- a/src/modules/referral/services/service.ts
+++ b/src/modules/referral/services/service.ts
@@ -171,18 +171,29 @@ export class ReferralService {
   }
 
   public subtractFunctionArgsFromCallData(data: string) {
-    const coder = new ethers.utils.AbiCoder();
-    // strip hex method identifier
-    const dataNoMethod = ethers.utils.hexDataSlice(data, 4);
-    // keep method hex identifier
-    const methodHex = data.replace(dataNoMethod.replace("0x", ""), "");
-    const decodedData = coder.decode(
-      ["address", "address", "uint256", "uint256", "uint64", "uint32"],
-      ethers.utils.hexDataSlice(data, 4),
-    );
-    const encoded = coder.encode(["address", "address", "uint256", "uint256", "uint64", "uint32"], decodedData);
-    const fullEncoded = methodHex + encoded.replace("0x", "");
-    return data.replace(fullEncoded, "");
+    try {
+      const coder = new ethers.utils.AbiCoder();
+      // strip hex method identifier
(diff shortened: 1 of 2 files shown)


dohaki (approved):
LGTM

src/modules/referral/services/service.ts
```diff
@@ -186,7 +186,13 @@ export class ReferralService {
   }
 
   public extractReferralAddressUsingDelimiter(data: string) {
-    const referralData = this.subtractFunctionArgsFromCallData(data);
+    let referralData = "";
+
+    try {
+      referralData = this.subtractFunctionArgsFromCallData(data);
```
nicholaspai, line 192: Do you think it'd be better to explicitly check the `data`'s length rather than the more vague try-catch which might catch a different error that we don't want to catch?

nicholaspai (commented):
https://github.com/across-protocol/scraper-api/pull/195/files#r1066039038

src/modules/referral/services/service.ts
amateima: I don't think we can check the data's length because it can be any length, right? You can attach any arbitrary data to the calldata

amateima:
> https://github.com/across-protocol/scraper-api/pull/195/files#r1066039038

@nicholaspai 
Modified the implementation a little bit to handle only the specific error

dohaki (dismissed):
LGTM

src/modules/referral/services/service.ts
```diff
@@ -170,19 +170,33 @@ export class ReferralService {
...
-    return data.replace(fullEncoded, "");
+    try {
+      const coder = new ethers.utils.AbiCoder();
+      // strip hex method identifier
+      const dataNoMethod = ethers.utils.hexDataSlice(data, 4);
+      // keep method hex identifier
+      const methodHex = data.replace(dataNoMethod.replace("0x", ""), "");
+      const decodedData = coder.decode(
+        ["address", "address", "uint256", "uint256", "uint64", "uint32"],
+        ethers.utils.hexDataSlice(data, 4),
```
pxrl, line 186: Could you use dataNoMethod here, instead of recreating the slice?

pxrl (approved):
Very clean and simple :+1:.

nicholaspai (approved):

PR #201, opened 2023-01-11
fix: don't retry consumer if deposit doesn't exist

amateima:
`BlockNumberConsumer` is a consumer attached to `BlockNumber` queue. It receives a deposit id and computes the deposit time based on the block number of the deposit. Currently, if the deposit doesn't exist in the DB (incorrect id), the job is retried infinitely. This can happen only if messages are published manually on this queue.


pxrl (approved):
Nice description!

dohaki (approved):
LGTM

PR #202, opened 2023-01-11
fix: don't retry consumer if deposit doesn't exist (prod)

amateima:
Production deployment of #201. See description there


dohaki (approved):
LGTM

PR #208, opened 2023-01-14
fix: improve query params sanitization

amateima:
This PR updates validation of the query params in order to restrict `limit` and `offset` within a certain range

### DIFF
--- a/src/modules/deposit/dto.ts
+++ b/src/modules/deposit/dto.ts
@@ -1,5 +1,6 @@
 import { ApiProperty } from "@nestjs/swagger";
-import { IsEnum, IsNumberString, IsOptional, IsString } from "class-validator";
+import { Type } from "class-transformer";
+import { IsEnum, IsInt, IsOptional, IsString, Max, Min } from "class-validator";
 
 export class GetDepositsQuery {
   @IsOptional()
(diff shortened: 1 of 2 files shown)


src/modules/deposit/dto.ts
```diff
@@ -16,13 +17,19 @@ export class GetDepositsQuery {
   status: "filled" | "pending";
 
   @IsOptional()
-  @IsNumberString({ no_symbols: true })
-  @ApiProperty({ example: "10", required: false })
+  @IsInt()
+  @Min(1)
+  @Max(100)
```
amateima, line 22: Restrict `?limit` to max 100. The FE client doesn't request more than 50 deposits per table page

src/modules/referral/entry-points/http/dto.ts
```diff
@@ -14,12 +15,18 @@ export class GetReferralsQuery {
   @ApiProperty({ example: "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D", minLength: 42, maxLength: 42, required: true })
   address: string;
 
-  @IsNumberString()
-  @ApiProperty({ example: "10", required: false })
+  @IsInt()
+  @Min(1)
+  @Max(100)
```
amateima, line 20: Restrict `?limit` to max 100. The FE client doesn't request more than 50 deposits per table page

dohaki (approved):
lgtm

pxrl (approved):

src/modules/deposit/dto.ts
```diff
@@ -16,13 +17,19 @@ export class GetDepositsQuery {
...
+  @Type(() => Number)
+  @ApiProperty({ example: 10, required: false })
   limit: string;
 
   @IsOptional()
-  @IsNumberString({ no_symbols: true })
-  @ApiProperty({ example: "0", required: false })
+  @IsInt()
+  @Min(0)
+  @Max(10_000_000)
```
mrice32, line 30: Does a large offset cause any problems? Would 10MM be significantly more expensive than 0?

mrice32 (commented):
Looks great! One comment

src/modules/deposit/dto.ts
amateima: It impacts the time execution of the query, yes

PR #211, opened 2023-01-25
refactor: rename deposit fill event and props amplitude (#209)

dohaki:
Update prod

Fixes ACX-674


dohaki:
> Looks good to me. It looks like there are multiple instances where we import packages but do not use them. This may affect our tree-shaking & could lead to degraded performance. We may want to restrict the eslint to prevent pushing code that has unused imports.

Yes we can remove them in a separate PR for cleanliness purposes. I think the performance gain via improved tree shaking in a nodejs environment are negligible though. Because we do not need to create a bundle like on the browser

PR #223, opened 2023-04-18
feat: improve the speed of querying contract events

amateima:


src/modules/scraper/service.ts
```diff
@@ -28,138 +28,97 @@ export class ScraperService {
...
+  public async publishBlocks(chainId: number, interval: number) {
     while (true) {
       try {
-        const latestBlocks = await this.getLatestBlocks();
-
-        if (this.appConfig.values.enableSpokePoolsEventsProcessing) {
-          await this.publishBlocks(latestBlocks);
-        }
-        if (this.appConfig.values.enableMerkleDistributorEventsProcessing) {
-          await this.publishMerkleDistributorBlocks(latestBlocks);
```
dohaki, line 39: We don't need to handle these conditions anymore? `this.appConfig.values.enableSpokePoolsEventsProcessing` and `this.appConfig.values.enableMerkleDistributorEventsProcessing`?

dohaki (approved):
Looks good 👍 Left one question and a nit

src/modules/scraper/service.ts
```diff
@@ -28,138 +28,97 @@ export class ScraperService {
...
 
   public async run() {
+    this.publishBlocks(ChainIds.mainnet, 10);
+    this.publishBlocks(ChainIds.arbitrum, 4);
+    this.publishBlocks(ChainIds.optimism, 4);
+    this.publishBlocks(ChainIds.polygon, 6);
+    this.publishMerkleDistributorBlocks(30);
+  }
+
+  public async publishBlocks(chainId: number, interval: number) {
```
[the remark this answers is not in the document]
mrice32: +1

src/modules/scraper/service.ts
```diff
@@ -28,138 +28,97 @@ export class ScraperService {
...
-          await this.publishMerkleDistributorBlocks(latestBlocks);
+        const blockNumber = await this.providers.getProvider(chainId).getBlockNumber();
+        this.logger.log(`latest block chainId: ${chainId} ${blockNumber}`);
+        const range = await this.determineBlockRange(
+          chainId,
+          blockNumber,
+          this.appConfig.values.web3.spokePoolContracts[chainId].startBlockNumber,
+          this.processedBlockRepository,
+        );
+        if (range) {
```
mrice32, line 49: A small comment either here or on determineBlockRange to explain when a caller would expect null/undefined would be helpful. My understanding from the code is that it only happens when from > to, right, which would mean we haven't moved forward since the last run?

src/modules/scraper/service.ts
```diff
@@ -28,138 +28,97 @@ export class ScraperService {
   }
 
   public async run() {
+    this.publishBlocks(ChainIds.mainnet, 10);
+    this.publishBlocks(ChainIds.arbitrum, 4);
+    this.publishBlocks(ChainIds.optimism, 4);
+    this.publishBlocks(ChainIds.polygon, 6);
```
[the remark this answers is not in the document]
amateima: Yes. Adding a comment now

src/modules/scraper/service.ts
amateima: yes. Basically it is either returning undefined or throwing an error. Anyway I added a comment for now

src/modules/scraper/service.ts
amateima: Good point 👍

src/modules/scraper/service.ts
```diff
@@ -28,138 +28,104 @@ export class ScraperService {
...
-    );
-
-    for (const chainId of Object.keys(blockRanges)) {
-      const { from, to } = blockRanges[chainId];
-      await this.scraperQueuesService.publishMessage<BlocksEventsQueueMessage>(ScraperQueue.BlocksEvents, {
-        chainId: parseInt(chainId),
-        from,
-        to,
-      });
+      await wait(secondsInterval);
```
james-a-morris, line 69: Does this block the entire thread?

src/modules/scraper/service.ts
```diff
@@ -28,138 +28,104 @@ export class ScraperService {
...
+        const configStartBlockNumber = this.appConfig.values.web3.merkleDistributor.startBlockNumber;
+        const range = await this.determineBlockRange(
+          chainId,
+          blockNumber,
+          configStartBlockNumber,
+          this.merkleDistributorProcessedBlockRepository,
+          true,
+        );
+
+        if (range) {
```
james-a-morris, line 87: Nit: should we explicitly test if the range is undefined/null or collapse it into a boolean with `!!range`?

james-a-morris (commented):
Left a few comments

src/modules/scraper/service.ts
amateima: the event loop you mean? no

src/modules/scraper/service.ts
amateima: Did it [here](https://github.com/across-protocol/scraper-api/pull/223/commits/4cd696bf06bf2c6e99603c713b8633a2e85be1a1)

src/modules/scraper/service.ts
```diff
@@ -28,138 +28,104 @@ export class ScraperService {
...
-
   public async determineBlockRange(
     chainId: number,
     latestBlockNumber: number,
-    configStartBlockNumber: number,
+    startBlockNumber: number,
     blockRepository: Repository<ProcessedBlock | MerkleDistributorProcessedBlock>,
+    keepDistanceFromHead = false,
   ) {
     let previousProcessedBlock = await blockRepository.findOne({
```
mrice32, line 115: I know this isn't changed, but how do you know the previous iteration has finished being processed before the next one starts? Is there any synchronization to ensure that you don't rerun the same range a second time because the prev one hasn't finished?

mrice32 (approved):
LGTM! Just one question

src/modules/scraper/service.ts
amateima: > how do you know the previous iteration has finished being processed before the next one starts

You don't know. This particular logic just defines the block ranges (`from`: last block registered in the DB; `to`: latest block number from chain) and submits it to the `BlocksEventsQueue`. The consumer attached to this queue will extract the events from that particular block range and the order is not mandatory to be maintained. The block ranges can be processed in any order

PR #232, opened 2023-05-04
feat: add support for SNX to cg service

dohaki:
Closes ACX-1067

### DIFF
--- a/src/modules/market-price/adapters/coingecko/index.ts
+++ b/src/modules/market-price/adapters/coingecko/index.ts
@@ -15,6 +15,7 @@ const symbolIdMap = {
   bal: "balancer",
   usdt: "tether",
   acx: "across-protocol",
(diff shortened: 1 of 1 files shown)


src/modules/market-price/adapters/coingecko/index.ts
```diff
@@ -15,6 +15,7 @@ const symbolIdMap = {
   bal: "balancer",
   usdt: "tether",
   acx: "across-protocol",
+  snx: "havven",
```
amateima, line 18: What a weird id

amateima (approved):
👍

src/modules/market-price/adapters/coingecko/index.ts
dohaki: Yea... I had to double-check because this seemed wrong to me

PR #236, opened 2023-05-11
feat: add hubpool scraper for root bundle events

james-a-morris:


james-a-morris:
We're no longer using the scraper for this use-case

PR #239, opened 2023-06-02
fix: add SpokePool2.5 contracts

dohaki:
I noticed that this branch was not merged to `stage` and only to `master`. After merging this one, we should sync both branches again

### DIFF
--- a/src/modules/scraper/utils.ts
+++ b/src/modules/scraper/utils.ts
@@ -24,8 +24,17 @@ export function splitBlockRanges(
           address: contracts[i].address,
           acrossVersion: contracts[i].acrossVersion,
         });
-      } else {
-        continue;
+      } else if (
+        contracts[i].startBlockNumber <= from &&
+        contracts[i + 1].startBlockNumber <= from &&
+        i === contracts.length - 2
+      ) {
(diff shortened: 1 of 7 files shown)


src/modules/scraper/utils.ts
```diff
@@ -24,8 +24,17 @@ export function splitBlockRanges(
           address: contracts[i].address,
           acrossVersion: contracts[i].acrossVersion,
         });
-      } else {
-        continue;
+      } else if (
+        contracts[i].startBlockNumber <= from &&
+        contracts[i + 1].startBlockNumber <= from &&
+        i === contracts.length - 2
```
james-a-morris, line 30: Should we place this earlier in the `else if` condition?

james-a-morris (approved):
LGTM - left a nit

amateima (approved):
We can also reset stage to master and force push, right? afaik there's no work on the staging branch that needs to be pushed to master

dohaki:
> We can also reset stage to master and force push, right? afaik there's no work on the staging branch that needs to be pushed to master

@amateima Yes, I think so too. I will close this PR then and reset `stage` to `master`.

PR #240, opened 2023-06-13
fix: add count when retry jobs

amateima:
Add an optional `count` parameter to specify how many queue jobs needs to be retried when calling the `scraper/failed-jobs/retry` endpoint.

**Issue**
With the current implementation, if the queue has many failed messages that need to be retried, the NodeJS app runs out of memory


dohaki (approved):
LGTM

PR #258, opened 2023-09-18
fix: fix sticky referral address query

amateima:
Fix the query that updates the sticky referral address of the deposits made by a particular depositor address.

### DIFF
--- a/src/modules/referral/services/queries.ts
+++ b/src/modules/referral/services/queries.ts
@@ -117,7 +117,8 @@ export const updateStickyReferralAddressesForDepositor = () => {
             order by d2."depositDate" desc
             limit 1
             ) d3 on d1."depositorAddr" = d3."depositorAddr"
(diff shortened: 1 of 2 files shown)


src/modules/referral/services/queries.ts
```diff
@@ -117,7 +117,8 @@ export const updateStickyReferralAddressesForDepositor = () => {
             order by d2."depositDate" desc
             limit 1
             ) d3 on d1."depositorAddr" = d3."depositorAddr"
-        where d1."depositorAddr" = $1
+        where d1."depositorAddr" = $1 
+          and d1."depositDate" >= $2
```
amateima, line 121: When a deposit `d` is processed by the DepositReferralConsumer to compute the sticky referral address, update the `stickyReferralAddress` only for the deposits with a `depositDate` >= `d`

PR #259, opened 2023-09-28
fix: move logic from sql query to js

amateima:
This PR changes the logic for computing the sticky referral address of deposits by moving away from making the computations using a single SQL query to using more JS logic.


amateima:
Still working on a new commit containing tests

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -39,6 +41,100 @@ export class DepositReferralConsumer {
...
+          },
+          order: {
+            depositDate: "DESC",
+          },
+        });
+        await this.depositRepository.update(
+          { id: d.id },
+          { stickyReferralAddress: previousDepositWithReferralAddress.referralAddress },
+        );
+      }
```
dohaki, line 117: NIT: You think it might be worth wrapping those in a `Promise.all`?

dohaki (approved):
Left one nit! TBH the JS implementation is easier to understand than the SQL query for me 😄

PR #271, opened 2023-11-06
fix: separate endpoint for pending deposits

amateima:
Move the logic that returns the pending deposits in a separate endpoint `/deposits/pending` because this logic is quite particular (filter out deposits older than 1 day and unprofitable deposits) and it needs to be taken out from the `/deposits` endpoint which is an endpoint suitable for more general use case.

The endpoint change must be made in the FE too


dohaki (commented):
This will only be used on the FE for the /all-transactions page right? The pending deposit of a user should be retrieved via the old endpoint?

amateima:
> This will only be used on the FE for the /all-transactions page right? The pending deposit of a user should be retrieved via the old endpoint?

Exactly. But most probably that functionality will get its own endpoint too

dohaki (approved):

PR #272, opened 2023-11-06
fix: add new endpoint to serve deposits

amateima:
I added a new `GET /deposits-v2` endpoint which is used to serve deposits. This endpoint is intended to replace `/deposits` in time because the old endpoint contains specific logic for some particular filters: for example if deposits are filtered by the `pending` status, it returns only profitable deposits happening in the past 24hrs. This kind of specific logic needs to go to dedicated endpoints, while the new `/deposits-v2` endpoint allows only simple filtering and sorting operations

### DIFF
--- a/src/modules/deposit/entry-point/http/controller.ts
+++ b/src/modules/deposit/entry-point/http/controller.ts
@@ -20,7 +20,7 @@ export class DepositController {
     return this.depositService.getDeposits(query.status, limit, offset);
   }
 
-  @Get("deposits-v2")
+  @Get("v2/deposits")
   getDepositsV2(@Query() query: GetDepositsV2Query) {
     return this.depositService.getDepositsV2(query);
   }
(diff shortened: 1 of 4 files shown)


src/modules/deposit/entry-point/http/controller.ts
```diff
@@ -20,6 +20,11 @@ export class DepositController {
     return this.depositService.getDeposits(query.status, limit, offset);
   }
 
+  @Get("deposits-v2")
```
dohaki, line 23: What do you think of using `v2/deposits` instead? I think it is generally best practice to version APIs like that

dohaki (commented):
Code looks good. Left one suggestion regarding versioning

src/modules/deposit/entry-point/http/controller.ts
amateima: Yeah, I hesitated to use `v2/deposits` because this implies that the whole API got a new version, but in fact it is only about upgrading a specific endpoint. I made the change tho 👍 It doesn't matter that much anyway because the plan is to replace the old /deposits endpoint with this one

dohaki (approved):
LGTM! Approving but e2e test shoud be fixed

PR #274, opened 2023-11-08
fix: add endpoint for referral rate

amateima:
This PR adds the `fields` query parameter to the `/referrals/summary` endpoint which allows to fetch only a subset of the fields returned by the endpoint.  

`GET /referrals/summary?address=${referrer}&fields[]=referralRate&fields[]=tier` endpoint returns only the referral rate and the referral tier


dohaki (approved):
LGTM

james-a-morris (commented):
Nothing actionable but LGTM pending @dohaki

PR #276, opened 2023-11-10
feat: index fee breakdown

dohaki:
@amateima I am targeting `stage` because I want to test this more in a proper environment. Should I first sync with master for that?


amateima:
Yes. Sync with master first

src/modules/deposit/model/deposit.entity.ts
```diff
@@ -37,6 +37,23 @@ export type RequestedSpeedUpDepositTx = {
   updatedRecipient?: string;
   updatedMessage?: string;
 };
+export type FeeBreakdown = {
+  // lp fee
+  lpFeeUsd: string;
+  lpFeePct: string; // wei pct
```
james-a-morris, line 43: We may not do it in other places, but should this be a NUMERIC()?

james-a-morris (commented):
left a few comments

src/modules/deposit/model/deposit.entity.ts
dohaki: As you can see [here](https://github.com/across-protocol/scraper-api/pull/276/files#diff-f9ec4a8d90d4d98e29761006de9633490a8df532a510b886b1c51713a8377c2aR149-R151), we are storing the fee breakdown as a `jsonb` type. I don't think it is possible to specify SQL specific types within `jsonb` 🤔

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
```diff
@@ -0,0 +1,42 @@
...
+      fillTxBlock.date,
+      destinationChainInfo.nativeSymbol.toLowerCase(),
+    );
+    const fee = new BigNumber(fillTxGasCostsWei).dividedBy(fixedPointAdjustment);
+
+    return {
+      fillTxBlockNumber: fillTxBlock.blockNumber,
+      fee: fee.toFixed(),
+      feeUsd: fee.multipliedBy(nativeTokenPriceUsd.usd).toFixed(),
+    };
```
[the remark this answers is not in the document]
dohaki: Yea, I think that's fine here. We don't need to be super precise of USD values

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
```diff
@@ -0,0 +1,42 @@
...
+      return {
+        fillTxBlockNumber: 0,
+        fee: "0",
+        feeUsd: "0",
+      };
+    }
+
+    const fillTxReceipt = await destinationChainProvider.getTransactionReceipt(fillTxHash);
+    // Some chains, e.g. Optimism, do not return the effective gas price in the receipt. We need to fetch it separately.
+    const gasPrice = fillTxReceipt.effectiveGasPrice || (await destinationChainProvider.getGasPrice());
```
amateima, line 27: But the gas price fetched separately will be different than the gas price at tx time, right? Is this ok?

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
```diff
@@ -0,0 +1,42 @@
...
+    const fillTxGasCostsWei = gasPrice.mul(fillTxReceipt.gasUsed).toString();
```
amateima, line 28: Is there a reason for which this naming is plural? "costs"

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
```diff
@@ -0,0 +1,42 @@
...
+    const fillTxBlock = await this.providers.getCachedBlock(destinationChainId, fillTxReceipt.blockNumber);
+    const nativeTokenPriceUsd = await this.marketPriceService.getCachedHistoricMarketPrice(
+      fillTxBlock.date,
+      destinationChainInfo.nativeSymbol.toLowerCase(),
+    );
```
amateima, line 33: I don't think this operation will succeed because I am not sure if `getCachedHistoricMarketPrice` will return prices for the current day, only starting with the day before. But we can test it.
If this is the case, then we should make sure that these types of edge cases are captured and errors are being thrown

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    if (!destinationChainProvider || !destinationChainInfo) {
+      return {
+        fillTxBlockNumber: 0,
+        fee: "0",
+        feeUsd: "0",
+      };
+    }
+
+    const fillTxReceipt = await destinationChainProvider.getTransactionReceipt(fillTxHash);
```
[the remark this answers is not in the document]
dohaki: Good call! Added here https://github.com/across-protocol/scraper-api/pull/276/commits/5122a949c216e39ab458dc3c44e4bd81d577abc3

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
```diff
@@ -0,0 +1,42 @@
...
+    const destinationChainProvider = this.providers.getProvider(destinationChainId);
+    const destinationChainInfo = chainIdToInfo[destinationChainId];
+
+    if (!destinationChainProvider || !destinationChainInfo) {
+      return {
+        fillTxBlockNumber: 0,
+        fee: "0",
+        feeUsd: "0",
+      };
+    }
```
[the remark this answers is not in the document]
dohaki: Yea makes sense. Changed here https://github.com/across-protocol/scraper-api/pull/276/commits/5122a949c216e39ab458dc3c44e4bd81d577abc3

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
dohaki: I think we can actually remove that fallback. I looked at the docs again of ethers and the library will actually use the normal `gasPrice` if there is no concept of a `baseFee`. See https://docs.ethers.org/v5/api/providers/types/#providers-TransactionReceipt

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
dohaki: Changed here https://github.com/across-protocol/scraper-api/commit/5122a949c216e39ab458dc3c44e4bd81d577abc3

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
dohaki: Yea, changed here https://github.com/across-protocol/scraper-api/commit/5122a949c216e39ab458dc3c44e4bd81d577abc3. If the API call fails, then the error should bubble up and this will throw and retry. If it returns `undefined` we now also throw and retry

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
```diff
@@ -0,0 +1,42 @@
+import { Injectable } from "@nestjs/common";
+import BigNumber from "bignumber.js";
+
+import { fixedPointAdjustment } from "../..//utils";
```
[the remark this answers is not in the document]
dohaki: https://github.com/across-protocol/scraper-api/commit/5122a949c216e39ab458dc3c44e4bd81d577abc3

amateima (approved):

PR #279, opened 2023-11-14
feat: new `/deposits/tx-page` endpoint for FE

dohaki:
During QA of the new transactions page here https://github.com/across-protocol/frontend-v2/pull/910, new requirements were:
- add more filters in the future
- and have a single deposits table that shows pending deposits first and then sorts by date

This PR adds these functionality to the new `/deposits/tx-page` endpoint.

### DIFF
--- a/src/modules/deposit/entry-point/http/controller.ts
+++ b/src/modules/deposit/entry-point/http/controller.ts
@@ -23,17 +23,27 @@ export class DepositController {
   @Get("deposits")
   @ApiTags("deposits")
   getDeposits(@Query() query: GetDepositsQuery) {
-    return this.depositService.getDepositsForTxPage({
-      userAddress: query.address || undefined,
-      status: query.status,
-      limit: query.limit,
-      offset: query.offset,
-    });
+    const limit = parseInt(query.limit ?? "10");
+    const offset = parseInt(query.offset ?? "0");
+
+    if (query.address) {
+      return this.depositService.getUserDeposits(query.address, query.status, limit, offset);
+    }
+
+    return this.depositService.getDeposits(query.status, limit, offset);
   }
 
   @Get("v2/deposits")
   getDepositsV2(@Query() query: GetDepositsV2Query) {
     return this.depositService.getDepositsV2(query);
+  }
+
+  @Get("deposits/pending")
+  @ApiTags("deposits")
+  getPendingDeposits(@Query() query: GetDepositsQuery) {
+    const limit = parseInt(query.limit ?? "10");
+    const offset = parseInt(query.offset ?? "0");
+    return this.depositService.getPendingDeposits(limit, offset);
   }
 
   @Get("deposits/details")
--- a/src/modules/deposit/entry-point/http/controller.ts
+++ b/src/modules/deposit/entry-point/http/controller.ts
@@ -1,23 +1,34 @@
 import { Controller, Get, Query } from "@nestjs/common";
 import { ApiResponse, ApiTags } from "@nestjs/swagger";
 import { DepositService } from "../../service";
-import { GetDepositsQuery, GetDepositDetailsQuery, GetDepositsStatsResponse, GetDepositsV2Query } from "./dto";
+import {
+  GetDepositsQuery,
+  GetDepositDetailsQuery,
+  GetDepositsStatsResponse,
+  GetDepositsV2Query,
+  GetDepositsForTxPageQuery,
+} from "./dto";
 
 @Controller()
 export class DepositController {
   public constructor(private depositService: DepositService) {}
 
+  @Get("deposits/tx-page")
+  @ApiTags("deposits")
+  getDepositsForTxPage(@Query() query: GetDepositsForTxPageQuery) {
+    return this.depositService.getDepositsForTxPage(query);
+  }
+
+  // TODO: Deprecate this endpoint if FE migrated to above `/deposits/tx-page` endpoint
   @Get("deposits")
   @ApiTags("deposits")
   getDeposits(@Query() query: GetDepositsQuery) {
-    const limit = parseInt(query.limit ?? "10");
-    const offset = parseInt(query.offset ?? "0");
-
-    if (query.address) {
-      return this.depositService.getUserDeposits(query.address, query.status, limit, offset);
-    }
-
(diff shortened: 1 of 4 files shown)


src/modules/deposit/entry-point/http/controller.ts
```diff
@@ -37,12 +48,4 @@ export class DepositController {
...
     return this.depositService.getCachedGeneralStats();
   }
-
-  @Get("deposits/pending")
-  @ApiTags("deposits")
-  getPendingDeposits(@Query() query: GetDepositsQuery) {
-    const limit = parseInt(query.limit ?? "10");
-    const offset = parseInt(query.offset ?? "0");
-    return this.depositService.getPendingDeposits(limit, offset);
-  }
```
dohaki, line 47: @amateima I recall that this endpoint was intended to be used on the FE. With the new `/deposits/tx-page` endpoint, I assume that this not needed anymore

src/modules/deposit/entry-point/http/dto.ts
```diff
@@ -130,6 +130,32 @@ export class GetDepositsV2Query {
...
+
+  @IsOptional()
+  @IsEnum(
+    {
+      STATUS: "status",
+    },
+    {
+      message: "Must be one of: 'status'",
+    },
+  )
```
james-a-morris, line 158: QQ: where do we specify the enum?

src/modules/deposit/service.ts
```diff
@@ -45,117 +45,64 @@ export class DepositService {
...
+  public async getDepositsForTxPage(query: Partial<GetDepositsForTxPageQuery>) {
     const limit = parseInt(query.limit ?? "10");
     const offset = parseInt(query.offset ?? "0");
-    let queryBuilder = this.depositRepository.createQueryBuilder("d");
-    queryBuilder = queryBuilder.where("d.depositDate is not null");
 
-    if (query.status) {
-      queryBuilder = queryBuilder.andWhere("d.status = :status", { status: query.status });
-    }
+    let queryBuilder = this.depositRepository.createQueryBuilder("d");
```
james-a-morris, line 52: QQ: what does `d` represent in this case?

src/modules/deposit/entry-point/http/dto.ts
dohaki: It gets inferred by the `@IsEnum` decorator AFAICT

src/modules/deposit/service.ts
dohaki: You can pass in an alias, in this case for the `Deposit` entity

src/modules/deposit/entry-point/http/controller.ts
```diff
@@ -1,23 +1,34 @@
...
-      return this.depositService.getUserDeposits(query.address, query.status, limit, offset);
-    }
-
-    return this.depositService.getDeposits(query.status, limit, offset);
+    return this.depositService.getDepositsForTxPage({
+      userAddress: query.address || undefined,
+      status: query.status,
+      limit: query.limit,
+      offset: query.offset,
+    });
```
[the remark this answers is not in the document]
dohaki: Yea, it's possible. The reasoning for reusing the `getDepositsForTxPage` method here was:
- Exactly the same behavior
- Returns the same formatted deposits
- Only used by our FE
- Reduce duplicated code
- Interface of the `/deposits` endpoint doesn't change

My question would be: Are you only concerned regarding the scope of this PR or would you never touch the existing endpoints in the future?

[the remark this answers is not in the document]
dohaki: Reverted [here](https://github.com/across-protocol/scraper-api/pull/279/commits/267f35a935140f405bd386941d03fab77f0618fc). It's just that we now have three methods with overlapping/duplicated code
- getPendingDeposits
- getDeposits
- getUserDeposits

that can be replaced with the new `getDepositsForTxPage`

src/modules/deposit/service.ts
james-a-morris: That feels like it degrades readability, but if it's used elsewhere sgtm!

src/modules/deposit/entry-point/http/dto.ts
james-a-morris: Right, but where do we define the enum in the code? Unless the enum is defined in-line with:
```
   {
      STATUS: "status",
    },
```

james-a-morris (commented):
resolved my open comments - as mentioned I don't have much to add but I will defer to @amateima

src/modules/deposit/service.ts
```diff
@@ -222,6 +272,33 @@ export class DepositService {
       acx_rewards_amount_referee: d.acxRewardsAmountReferee,
     }));
   }
+
+  private _getFilteredDepositsQuery(
```
amateima, line 276: I would not use the _ prefix since typescript already have a dedicated keyword for private

src/modules/deposit/service.ts
```diff
@@ -132,30 +202,10 @@ export class DepositService {
...
-
-    if (query.tokenAddress) {
-      queryBuilder = queryBuilder.andWhere("d.tokenAddr = :tokenAddr", { tokenAddr: query.tokenAddress });
-    }
+    let queryBuilder = this.depositRepository.createQueryBuilder("d");
+    queryBuilder = this._getFilteredDepositsQuery(queryBuilder, query);
 
-    queryBuilder = queryBuilder.orderBy("d.depositDate", "DESC");
-    queryBuilder = queryBuilder.take(limit);
-    queryBuilder = queryBuilder.skip(offset);
```
amateima, line 158: These 3 should not be removed, right?

amateima (changes requested):

src/modules/deposit/service.ts
dohaki: https://github.com/across-protocol/scraper-api/pull/279/commits/a44ac9d35236845a1fc0a7d5ac196bb24c250ac3

src/modules/deposit/service.ts
dohaki: https://github.com/across-protocol/scraper-api/pull/279/commits/a44ac9d35236845a1fc0a7d5ac196bb24c250ac3

PR #284, opened 2023-11-15
fix: refactor referral address mechanism

amateima:
This is a refactoring of the mechanism used for extracting the referral address and computing the sticky referral address

**How it works**
1. A message containing a `depositId` is submitted on the `DepositReferral` queue and it is processed by [DepositReferralConsumer](https://github.com/across-protocol/scraper-api/blob/9d7e48c26723c680e58a0719c698d337fce03fcf/src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts#L25-L55). This consumer has only 1 job: it extracts the referral address from the tx data or it derives the sticky referral address from previous deposits made by the same depositor
2. The consumer from above also publishes a message on the `RectifyStickyReferral` queue and the message is processed by [RectifyStickyReferralConsumer](https://github.com/across-protocol/scraper-api/blob/9d7e48c26723c680e58a0719c698d337fce03fcf/src/modules/scraper/adapter/messaging/RectifyStickyReferralConsumer.ts#L21-L41). If there are deposits in the DB with a sticky referral address, but an earlier deposit with a referral address is inserted in the DB at a later time, then the sticky referral address needs to be recomputed. This is achieved by detecting such deposits and re-publish then on the `DepositReferral` queue

**New API endpoints**

There is a new admin [endpoint](https://github.com/across-protocol/scraper-api/pull/284/files#diff-6584665fe6b6350a2fbd5484d734bd1ff0cf61d68dab6699a06c4f8e16fa95d4R126-R134) `POST /scraper/referral-address-reindex` which [publishes](https://github.com/across-protocol/scraper-api/pull/284/files#diff-bb8f574ca10ada36c4348f8454670dbfb452ba295bfafe7e0ca187b893590a8bR219-R253) all the deposits within a particular date range to the `DepositReferral` queue (step 1 from above). This endpoint is necessary because we need to recompute all the sticky referral addresses for October because of a bug that caused the sticky referral address to be computed incorrectly for a few deposits

### DIFF
--- a/src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
@@ -1,19 +1,57 @@
 import { OnQueueFailed, Process, Processor } from "@nestjs/bull";
 import { Logger } from "@nestjs/common";
 import { Job } from "bull";
-import { DepositReferralQueueMessage, ScraperQueue } from ".";
+import { Repository } from "typeorm";
+import { InjectRepository } from "@nestjs/typeorm";
+
+import { DepositReferralQueueMessage, RectifyStickyReferralQueueMessage, ScraperQueue } from ".";
 import { ReferralService } from "../../../referral/services/service";
+import { Deposit } from "../../../deposit/model/deposit.entity";
+import { EthProvidersService } from "../../../web3/services/EthProvidersService";
+import { ScraperQueuesService } from "../../service/ScraperQueuesService";
 
 @Processor(ScraperQueue.DepositReferral)
 export class DepositReferralConsumer {
   private logger = new Logger(DepositReferralConsumer.name);
 
-  constructor(private referralService: ReferralService) {}
+  constructor(
+    private referralService: ReferralService,
+    @InjectRepository(Deposit) readonly depositRepository: Repository<Deposit>,
+    private ethProvidersService: EthProvidersService,
+    private scraperQueuesService: ScraperQueuesService,
+  ) {}
 
   @Process({ concurrency: 1 })
   private async process(job: Job<DepositReferralQueueMessage>) {
-    const { depositId } = job.data;
-    await this.referralService.extractReferralAddressAndComputeStickyReferralAddresses(depositId);
+    const { depositId, rectifyStickyReferralAddress } = job.data;
+    const deposit = await this.depositRepository.findOne({ where: { id: depositId } });
+
+    if (!deposit) return;
+    // if depositDate field is missing, throw an error to retry the message as this field
+    // is necessary to compute the sticky referral address.
+    if (!deposit.depositDate) throw new Error(`depositId ${deposit.id}: wait for depositDate`);
+
+    const { depositTxHash, sourceChainId } = deposit;
+    const transaction = await this.ethProvidersService.getCachedTransaction(sourceChainId, depositTxHash);
+    const block = await this.ethProvidersService.getCachedBlock(sourceChainId, transaction.blockNumber);
+    const blockTimestamp = parseInt((new Date(block.date).getTime() / 1000).toFixed(0));
+
+    if (!transaction) throw new Error("Transaction not found");
+
+    await this.referralService.extractReferralAddressOrComputeStickyReferralAddresses({
+      blockTimestamp,
+      deposit,
+      transactionData: transaction.data,
+    });
+
+    if (rectifyStickyReferralAddress) {
+      await this.scraperQueuesService.publishMessage<RectifyStickyReferralQueueMessage>(
+        ScraperQueue.RectifyStickyReferral,
+        {
+          depositId: deposit.id,
+        },
+      );
+    }
   }
 
   @OnQueueFailed()
--- a/src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
@@ -38,13 +38,13 @@ export class DepositReferralConsumer {
 
     if (!transaction) throw new Error("Transaction not found");
 
-    await this.referralService.extractReferralAddressOrComputeStickyReferralAddresses({
+    const { referralAddress } = await this.referralService.extractReferralAddressOrComputeStickyReferralAddresses({
       blockTimestamp,
       deposit,
       transactionData: transaction.data,
     });
 
-    if (rectifyStickyReferralAddress) {
+    if (rectifyStickyReferralAddress && referralAddress) {
       await this.scraperQueuesService.publishMessage<RectifyStickyReferralQueueMessage>(
         ScraperQueue.RectifyStickyReferral,
         {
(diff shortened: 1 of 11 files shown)


src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -1,19 +1,57 @@
...
+    // if depositDate field is missing, throw an error to retry the message as this field
+    // is necessary to compute the sticky referral address.
+    if (!deposit.depositDate) throw new Error(`depositId ${deposit.id}: wait for depositDate`);
+
+    const { depositTxHash, sourceChainId } = deposit;
+    const transaction = await this.ethProvidersService.getCachedTransaction(sourceChainId, depositTxHash);
+    const block = await this.ethProvidersService.getCachedBlock(sourceChainId, transaction.blockNumber);
+    const blockTimestamp = parseInt((new Date(block.date).getTime() / 1000).toFixed(0));
+
+    if (!transaction) throw new Error("Transaction not found");
```
dohaki, line 39: Is retrying the message in this case the wanted behavior? If a transaction is not found then
1. Either the tx wasn't indexed or mined
2. Or a tx with the hash doesn't exist

Case 1 can be excluded because the deposit was stored through an emitted contract event. And if case 2, then this should just return, no?

src/modules/scraper/adapter/messaging/RectifyStickyReferralConsumer.ts
```diff
@@ -0,0 +1,47 @@
+import { OnQueueFailed, Process, Processor } from "@nestjs/bull";
+import { Logger } from "@nestjs/common";
+import { Job } from "bull";
+import { IsNull, MoreThan, Not, Repository } from "typeorm";
```
dohaki, line 4:
```suggestion
import { MoreThan, Repository } from "typeorm";
```

src/modules/scraper/adapter/messaging/RectifyStickyReferralConsumer.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (!deposit) return;
+
+    // After extracting the referral address, all deposits with later deposit time must have
+    // the sticky referral address updated
+    const deposits = await this.depositRepository.find({
+      where: {
+        depositorAddr: deposit.depositorAddr,
+        depositDate: MoreThan(deposit.depositDate),
+      },
+    });
```
dohaki, line 34: Do we need to sort these by the deposit date or doesn't the order matter here?

dohaki (commented):
Left some comments and questions

src/modules/scraper/entry-point/http/dto.ts
```diff
@@ -36,6 +36,16 @@ export class SubmitReferralAddressJobBody {
   toDepositId: number;
 }
 
+export class SubmitReindexReferralAddressJobBody {
+  @IsDateString()
+  @ApiProperty({ example: "2022-11-08T11:00:00.000Z", required: true })
```
james-a-morris, line 41: This is for the swagger docs, right?

src/modules/scraper/adapter/messaging/RectifyStickyReferralConsumer.ts
```diff
@@ -0,0 +1,47 @@
...
+    @InjectRepository(Deposit) readonly depositRepository: Repository<Deposit>,
+    private scraperQueuesService: ScraperQueuesService,
+  ) {}
+
+  @Process({ concurrency: 1 })
+  private async process(job: Job<RectifyStickyReferralQueueMessage>) {
+    const { depositId } = job.data;
+    const deposit = await this.depositRepository.findOne({ where: { id: depositId } });
+
+    if (!deposit) return;
```
mrice32, line 25: Similar question: how would this ever happen? And why isn't it a problem?

src/modules/scraper/entry-point/http/controller.ts
```diff
@@ -117,10 +118,20 @@ export class ScraperController {
...
       });
     }
   }
 
+  @Post("scraper/referral-address-reindex")
+  @Roles(Role.Admin)
+  @UseGuards(JwtAuthGuard, RolesGuard)
+  @ApiBearerAuth()
+  @ApiTags("scraper")
+  async submitReindexReferralAddressJob(@Body() body: SubmitReindexReferralAddressJobBody) {
```
mrice32, line 131: The intention of this job is just to allow manual reindexing in the case that somebody reports an issue?

mrice32 (approved):
Looks great! Only minor questions and comments.

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
amateima: Yes, retrying is the wanted behaviour, because, as you said, this is very unlikely to happen, but if it happens, then it means that there's definitely a problem somewhere. So I prefer the message to be retried, so when the issue/bug is fixed, then the message will be retried successfully

src/modules/scraper/entry-point/http/controller.ts
amateima: Yes. It takes ALL the deposits within a timeframe (`fromDate` -> `toDate`) and it will re-compute the referral addresses OR the sticky referral addresses by submitting the deposits ids to the `DepositReferralConsumer`

src/modules/scraper/adapter/messaging/RectifyStickyReferralConsumer.ts
amateima: This will happen if an invalid deposit is manually submitted to the `RectifyStickyReferral` queue (for example, by using an admin endpoint). In this case we don't want to retry the message indefinitely

src/modules/scraper/adapter/messaging/DepositReferralConsumer.ts
```diff
@@ -1,19 +1,57 @@
...
+  ) {}
 
   @Process({ concurrency: 1 })
   private async process(job: Job<DepositReferralQueueMessage>) {
-    const { depositId } = job.data;
-    await this.referralService.extractReferralAddressAndComputeStickyReferralAddresses(depositId);
+    const { depositId, rectifyStickyReferralAddress } = job.data;
+    const deposit = await this.depositRepository.findOne({ where: { id: depositId } });
+
+    if (!deposit) return;
```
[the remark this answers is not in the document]
amateima: There is no way a deposit can be deleted from the DB.
This `if` instruction will pass only if an invalid deposit is manually submitted to the `DepositReferral` queue (for example, by using an admin endpoint). In this case we don't want to retry the message indefinitely

src/modules/scraper/entry-point/http/dto.ts
amateima: Exactly!

src/modules/scraper/adapter/messaging/RectifyStickyReferralConsumer.ts
amateima: Order doesn't matter here, but I added sorting by deposit date because it makes from a logical point of view. Thank you!

amateima:
ACX-1707

PR #293, opened 2023-11-21
feat: include rewards to /deposits/tx-page response

dohaki:
Closes ACX-1714

This adds a new `RewardService` in `src/modules/rewards/services/reward-service.ts` which serves as an umbrella service for all reward-related functionality. This service is used by the deposits module to include related rewards for a deposit.

Requires #286

### DIFF
--- a/src/modules/deposit/service.ts
+++ b/src/modules/deposit/service.ts
@@ -91,7 +91,7 @@ export class DepositService {
     const [deposits, total] = await queryBuilder.getManyAndCount();
 
     // Only include rewards if a user address is provided
-    if (query.include && query.depositorOrRecipientAddress && query.include.includes("rewards")) {
+    if (query.include && query.depositorOrRecipientAddress) {
       const userAddress = this.assertValidAddress(query.depositorOrRecipientAddress);
       const rewards = await this.rewardService.getRewardsForDepositsAndUserAddress(deposits, userAddress);
       const enrichedDeposits = this.rewardService.enrichDepositsWithRewards(deposits, rewards);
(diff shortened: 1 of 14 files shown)


src/modules/deposit/service.ts
```diff
@@ -304,34 +325,3 @@ export class DepositService {
     return queryBuilder;
   }
 }
-
-/**
- * Format deposit to entity to match `Transfer` structure from sdk.
- * @param deposit - Deposit entity from db.
- * @returns Formatted deposit entity that matches `Transfer` struct.
- */
-export function formatDeposit(deposit: Deposit) {
```
dohaki, line 313: Had to remove this in a separate `utils` file due to circular dependency path

src/modules/referral/services/service.ts
```diff
@@ -172,6 +170,33 @@ export class ReferralService {
...
+          ? 1
+          : item.depositorAddr === userAddress
+          ? 0.25
+          : 0.75,
+      )
+        .multipliedBy(item.referralRate)
+        .multipliedBy(item.multiplier);
+      const acxRewards = new BigNumber(item.bridgeFeeUsd)
+        .dividedBy(new BigNumber(item.acxUsdPrice).dividedBy(new BigNumber(10).pow(18)))
+        .multipliedBy(appliedRate);
```
dohaki, line 191: I wanted to reuse the existing `getReferralsQuery` in `src/modules/referral/services/queries.ts` but couldn't get the `IN (:...keys)` operator to work with params and raw SQL. So I used this workaround. Happy to change if another solution works better

amateima (approved):

PR #296, opened 2023-11-23
feat: serve asset decimals

dohaki:
Clique requested decimals


amateima (commented):
I have one suggestion. Let's add a new query parameter `include` which can be used to specify other entities that can be attached to the default response.
e.g `/v2/deposits?include[]=token&include[]=TBD` would make the join with the `token` table and add a new `token` parameter to the response. The added parameter to the response and its value would look like:
```
token: { symbol, decimals, address, name }
```

This implementation is very flexible so this way we provide a mechanism for adapting the response to many kind of situations: adding rewards to the response, add historical prices to the response, etc.

Wdyt?

dohaki:
Closing in favor of https://github.com/across-protocol/scraper-api/pull/297

PR #302, opened 2023-11-27
fix: fee breakdown zero values

dohaki:
The `FeeBreakdownConsumer` didn't correctly handle `0` values. If a slow fill, for example, occurs then the `relayerFeePct` is `0`.

### DIFF
--- a/src/modules/web3/model/tx-receipt.entity.ts
+++ b/src/modules/web3/model/tx-receipt.entity.ts
@@ -12,10 +12,10 @@ export class TransactionReceipt {
   @Column()
   from: string;
 
-  @Column()
+  @Column({ nullable: true })
   to: string;
 
-  @Column()
+  @Column({ nullable: true })
   contractAddress: string;
 
   @Column()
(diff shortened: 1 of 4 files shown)


src/modules/web3/model/tx-receipt.entity.ts
```diff
@@ -12,10 +12,10 @@ export class TransactionReceipt {
   @Column()
   from: string;
 
-  @Column()
+  @Column({ nullable: true })
   to: string;
```
amateima, line 16:
```suggestion
  to?: string;
```

src/modules/web3/model/tx-receipt.entity.ts
```diff
@@ -12,10 +12,10 @@ export class TransactionReceipt {
...
 
-  @Column()
+  @Column({ nullable: true })
   contractAddress: string;
```
amateima, line 19:
```suggestion
  contractAddress?: string;
```

amateima (approved):

PR #304, opened 2023-11-27
feat: admin endpoint for backfilling fee breakdowns

dohaki:
Additional convenience admin endpoint for backfilling fee breakdowns

### DIFF
--- a/src/modules/scraper/entry-point/http/controller.ts
+++ b/src/modules/scraper/entry-point/http/controller.ts
@@ -32,6 +32,7 @@ import {
   SubmitReindexReferralAddressJobBody,
   SubmitFeeBreakdownBody,
   OpRebateRewardBody,
+  BackfillFeeBreakdownBody,
 } from "./dto";
 
 @Controller()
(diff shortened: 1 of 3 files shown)


src/modules/scraper/entry-point/http/controller.ts
```diff
@@ -228,4 +229,13 @@ export class ScraperController {
...
     await this.scraperService.retryIncompleteDeposits(body);
   }
+
+  @Post("scraper/fee-breakdown/backfill")
+  @ApiTags("scraper")
+  @Roles(Role.Admin)
+  @UseGuards(JwtAuthGuard, RolesGuard)
+  @ApiBearerAuth()
+  async backfillFeeBreakdown(@Body() body: BackfillFeeBreakdownBody) {
+    await this.scraperService.backfillFeeBreakdown(body);
```
james-a-morris, line 239: Should we be returning something that has a status?

dohaki: If this method doesn't throw, then NestJS automatically returns a success status code 200/201

PR #319, opened 2023-12-11
fix: refine fee breakdown errors and enforce filled status

dohaki:
Not 100% sure if this will fix ACX-1741 but should at least give a bit more insight in the logs I hope.

### DIFF
--- a/src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
+++ b/src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
@@ -21,7 +21,7 @@ export class GasFeesService {
     const fillTxGasCostWei = new BigNumber(fillTxReceipt.effectiveGasPrice).multipliedBy(fillTxReceipt.gasUsed);
     const fillTxBlock = await this.providers.getCachedBlock(destinationChainId, fillTxReceipt.blockNumber);
     const nativeTokenPriceUsd = await this.marketPriceService.getCachedHistoricMarketPrice(
-      new Date(fillTxBlock.date),
+      fillTxBlock.date,
       destinationChainInfo.nativeSymbol.toLowerCase(),
     );
     const fee = fillTxGasCostWei.dividedBy(fixedPointAdjustment);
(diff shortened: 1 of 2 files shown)


src/modules/scraper/adapter/messaging/FeeBreakdownConsumer.ts
```diff
@@ -31,8 +31,16 @@ export class FeeBreakdownConsumer {
...
 
-    if (!deposit.token || !deposit.price || !deposit.depositDate) {
-      throw new Error("Can not get fee breakdown without token, price or deposit date");
+    if (!deposit.token) {
+      throw new Error("Token not populated");
+    }
+
+    if (!deposit.price) {
+      throw new Error("Price not populated");
+    }
```
james-a-morris, line 40: Do we have access to the SDK? If so, can we use `isDefined()`

james-a-morris (commented):
Left a comment

src/modules/scraper/adapter/gas-fees/gas-fees-service.ts
```diff
@@ -21,7 +21,7 @@ export class GasFeesService {
     const fillTxGasCostWei = new BigNumber(fillTxReceipt.effectiveGasPrice).multipliedBy(fillTxReceipt.gasUsed);
     const fillTxBlock = await this.providers.getCachedBlock(destinationChainId, fillTxReceipt.blockNumber);
     const nativeTokenPriceUsd = await this.marketPriceService.getCachedHistoricMarketPrice(
-      fillTxBlock.date,
+      new Date(fillTxBlock.date),
```
amateima, line 24: `fillTxBlock.date` is already a Date 🤔

dohaki: oh yea true. reverted this change

src/modules/scraper/adapter/messaging/FeeBreakdownConsumer.ts
dohaki: No, we currently don't use the SDK the Scraper. If we need it at some point, we could use the helper yea

amateima (approved):

PR #322, opened 2023-12-14
feat: add endpoint for backfilling filledDate

dohaki:
Admin endpoint for setting `filledDate`.

There are quite a lot of deposits where filled date is null even though they are filled. This is also causing a lot of errors in `TrackFillEventsConsumer`

### DIFF
--- a/src/modules/scraper/entry-point/http/dto.ts
+++ b/src/modules/scraper/entry-point/http/dto.ts
@@ -131,3 +131,10 @@ export class BackfillFeeBreakdownBody {
   @ApiProperty({ example: 0 })
   count?: number;
 }
+
+export class BackfillFilledDateBody {
+  @IsOptional()
+  @IsInt()
+  @ApiProperty({ example: 0 })
+  count?: number;
+}
--- a/src/modules/scraper/service.ts
+++ b/src/modules/scraper/service.ts
@@ -11,6 +11,7 @@ import { ScraperQueuesService } from "./service/ScraperQueuesService";
 import {
   BlockNumberQueueMessage,
   BlocksEventsQueueMessage,
+  DepositFilledDateQueueMessage,
   DepositReferralQueueMessage,
   FeeBreakdownQueueMessage,
   MerkleDistributorBlocksEventsQueueMessage,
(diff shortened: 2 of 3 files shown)


src/modules/scraper/service.ts
```diff
@@ -281,4 +283,23 @@ export class ScraperService {
...
+      .andWhere("d.filledDate is null")
+      .andWhere("d.status='filled'")
+      .andWhere("d.depositDate is not null")
+      .orderBy("d.id", "DESC")
+      .take(body.count || 1000)
+      .getMany();
+    this.logger.debug(`[backfillFilledDate] found ${deposits.length} deposits`);
+
+    for (const deposit of deposits) {
+      this.logger.debug(`[backfillFilledDate] get filled date for ${deposit.id}`);
```
amateima, line 299: Do you want to keep this log?

amateima (approved):

src/modules/scraper/entry-point/http/dto.ts
```diff
@@ -131,3 +131,10 @@ export class BackfillFeeBreakdownBody {
   @ApiProperty({ example: 0 })
   count?: number;
 }
+
+export class BackfillFilledDateBody {
+  @IsOptional()
+  @IsInt()
+  @ApiProperty({ example: 0 })
```
james-a-morris, line 138: Could we also include an example of this being undefined?

src/modules/scraper/service.ts
james-a-morris, line 299: This logging call is unbounded by the code. I'm not sure this is a good idea to log each one.

james-a-morris (commented):
Left a comment

src/modules/scraper/service.ts
dohaki: removed

src/modules/scraper/entry-point/http/dto.ts
dohaki: Hmm, why do you want to add an example of this being undefined? 🤔

src/modules/scraper/service.ts
dohaki: Removed!

PR #333, opened 2024-01-09
feat: implement op rewards airdrop

amateima:


dohaki (changes requested):
Code looks good! One e2e test seems to fail though:
```
FAIL test/referrals.e2e-spec.ts (10.434 s)
  DELETE /referrals/merkle-distribution
    ✓ return 401 (29 ms)
    ✕ return 201 (1073 ms)
```
I think you need to fix the URLs in:
https://github.com/across-protocol/scraper-api/blob/3138993df5763a31b8fa75f98549b1ef7ac75b8d/test/referrals.e2e-spec.ts#L130
https://github.com/across-protocol/scraper-api/blob/3138993df5763a31b8fa75f98549b1ef7ac75b8d/test/referrals.e2e-spec.ts#L111

src/modules/referral/entry-points/http/dto.ts
```diff
@@ -1,6 +1,6 @@
 import { ApiProperty } from "@nestjs/swagger";
 import { Type } from "class-transformer";
-import { IsArray, IsDateString, IsInt, IsNumberString, IsOptional, IsString, Length, Max, Min } from "class-validator";
+import { IsArray, IsInt, IsNumberString, IsOptional, IsString, Length, Max, Min } from "class-validator";
```
dohaki, line 3: `IsNumberString` can be removed?

dohaki (approved):
LGTM! One nit

PR #334, opened 2024-01-10
feat: fetch proofs depending on rewards type

amateima:
Should be rebased on top of `master` after #333 is merged


dohaki (commented):
one test seems to fail

dohaki (approved):
Nice!

PR #336, opened 2024-01-31
fix: set depositor address correctly for deposits made through SpokePoolVerifier contract

amateima:
Set the depositor as being the transaction sender for deposits made through the SpokePoolVerifier contract

### DIFF
--- a/src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
@@ -36,6 +36,8 @@ import { AppConfig } from "../../../configuration/configuration.service";
 import { splitBlockRanges } from "../../utils";
 import { Event } from "ethers";
 
+const SPOKE_POOL_VERIFIER_CONTRACT_ADDRESS = "0x269727F088F16E1Aea52Cf5a97B1CD41DAA3f02D";
+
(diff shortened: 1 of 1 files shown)


src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
```diff
@@ -242,11 +244,18 @@ export class BlocksEventsConsumer {
...
+  private async fromFundsDepositedEventToDeposit(chainId: number, event: Event) {
     const typedEvent = event as FundsDepositedEvent2 | FundsDepositedEvent2_5;
     const { transactionHash, blockNumber } = typedEvent;
     const { depositId, originChainId, destinationChainId, amount, originToken, depositor, relayerFeePct, recipient } =
       event.args;
+    // In some cases the depositor is the txn msg.sender
+    let trueDepositor = depositor;
+
+    if (depositor === SPOKE_POOL_VERIFIER_CONTRACT_ADDRESS) {
+      const tx = await this.providers.getProvider(chainId).getTransaction(transactionHash);
```
dohaki, line 256: I believe that we already store the receipt in the db 🤔 Could be worth to use `getCachedTransactionReceipt`?

amateima: You are right! 👍

PR #353, opened 2024-02-22
fix: improve capped bridge fee pct

amateima:
This PRs [removes the computation of the capped bridge fee pct](https://github.com/across-protocol/scraper-api/pull/353/files#diff-ecfac74c90bff76d0ddbb76c32110ca565d1b91dfebb79fce0ea9f9ad8a3d175L54) from the consumer that process V3 fill events and moves it to its own dedicated & retryable consumer: [CappedBridgeFeeConsumer](https://github.com/across-protocol/scraper-api/pull/353/files#diff-5f582b450e40be10f254993aa396a690607a8296a2f0a84b256bf968f2da7694). 
The reason is simple: for V3 deposits, the computation of the bridge fee percentage needs the prices of the input & output token. Fetching prices is done asynchronously, in a different consumer,  so there's no guarantee that they are available by the time the fills are processed for a particular deposit.

⚠️  Also, the bridge fee pct is computed based on the USD input and output amounts, to avoid issues when the output and input token have different decimals number

### DIFF
--- a/src/modules/scraper/adapter/messaging/CappedBridgeFeeConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/CappedBridgeFeeConsumer.ts
@@ -7,7 +7,7 @@ import BigNumber from "bignumber.js";
 
 import { CappedBridgeFeeQueueMessage, ScraperQueue } from ".";
 import { Deposit } from "../../../deposit/model/deposit.entity";
-import { AcrossContractsVersion } from "src/modules/web3/model/across-version";
+import { AcrossContractsVersion } from "../../../web3/model/across-version";
 
 /**
  * This consumer computes the the capped bridge fee percentage used for computing referral rewards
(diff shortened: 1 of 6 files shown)


src/modules/scraper/adapter/messaging/CappedBridgeFeeConsumer.ts
```diff
@@ -0,0 +1,93 @@
...
+import { Repository } from "typeorm";
+import { InjectRepository } from "@nestjs/typeorm";
+import BigNumber from "bignumber.js";
+
+import { CappedBridgeFeeQueueMessage, ScraperQueue } from ".";
+import { Deposit } from "../../../deposit/model/deposit.entity";
+import { AcrossContractsVersion } from "../../../web3/model/across-version";
+
+/**
+ * This consumer computes the the capped bridge fee percentage used for computing referral rewards
```
dohaki, line 13:
```suggestion
 * This consumer computes the capped bridge fee percentage used for computing referral rewards
```

dohaki (approved):
LGTM

PR #361, opened 2024-03-07
feat(linea): add support for linea

dohaki:
Closes ACX-1776


amateima (commented):
Did we add the Linea providers in `zion`?

dohaki:
> Did we add the Linea providers in `zion`?

Not yet. I will do that before merging

PR #364, opened 2024-03-19
fix: cache total deposits count for tx page

amateima:
Cache the total number of deposits for the `/tx-page` endpoint for 5 minutes to avoid successive `COUNT` SQL queries


dohaki (approved):
LGTM

PR #390, opened 2024-05-02
fix: fee breakdown v3 events relay gas fee amount

dohaki:
I noticed wrong `relayGasFeeAmount` values for deposits where the token decimals != 18. E.g.
image

### DIFF
--- a/src/modules/scraper/adapter/messaging/FeeBreakdownConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/FeeBreakdownConsumer.ts
@@ -74,10 +74,6 @@ export class FeeBreakdownConsumer {
     const typedFillTx = deposit.fillTxs as DepositFillTxV3[];
     const fillTx = typedFillTx[0];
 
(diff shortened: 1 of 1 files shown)


src/modules/scraper/adapter/messaging/FeeBreakdownConsumer.ts
```diff
@@ -93,7 +99,7 @@ export class FeeBreakdownConsumer {
       relayCapitalFeePct: undefined,
       relayCapitalFeeAmount: undefined,
       relayGasFeeUsd,
-      relayGasFeePct: undefined,
+      relayGasFeePct: relayGasFeePct.toFixed(0),
```
amateima, line 102: Do you plan to use `relayGasFeePct` in the FE instead of `relayGasFeeAmount`?

amateima (approved):

src/modules/scraper/adapter/messaging/FeeBreakdownConsumer.ts
dohaki: Not sure yet... But I thought it would not hurt to add this

PR #405, opened 2024-05-13
feat: identify eth deposits

amateima:


src/modules/web3/model/weth-events.ts
```diff
@@ -0,0 +1,37 @@
...
+  };
+}
+
+export interface WethTransfetEventArbitrum extends Event {
+  args: [string, string, BigNumber] & {
+    from: string;
+    to: string;
+    value: BigNumber;
+  };
+}
```
dohaki, line 37: Any reason why we need duplicated events for different chains? AFAICT only on Arbitrum the respective event is different. But can't we just reuse the `Deposit` event signature for every other chain? This would also decrease the number of `else if` cases in the code

src/modules/web3/model/weth-events.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface WethDepositEventBase extends Event {
+  args: [string, BigNumber] & {
+    dst: string;
+    wad: BigNumber;
+  };
+}
+
+export interface WethTransfetEventArbitrum extends Event {
```
dohaki, line 31:
```suggestion
export interface WethTransferEventArbitrum extends Event {
```

src/modules/web3/services/abi/WETH_1.json
dohaki, line 1: I don't think we need a different ABI file for every chain if they are the same, no?

src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
```diff
@@ -204,52 +222,210 @@ export class BlocksEventsConsumer {
...
+              typedEvent.args.wad.eq(depositEvent.args.inputAmount) &&
+              spokePoolAddresses.includes(typedEvent.args.dst)
+            );
+          } else if (chainId === ChainIds.arbitrum) {
+            const typedEvent = e as WethTransfetEventArbitrum;
+            return (
+              e.logIndex < depositEvent.logIndex &&
+              typedEvent.args.value.eq(depositEvent.args.inputAmount) &&
+              spokePoolAddresses.includes(typedEvent.args.to)
+            );
```
dohaki, line 295: Don't we need to also add the check
```
typedEvent.args.from === ZERO_ADDRESS
```
? So that we don't consider a normal WETH transfer but rather a "deposit", i.e. mint.

dohaki (commented):
Left a few questions and remarks

src/modules/web3/services/abi/WETH_1.json
amateima: Unfortunately the ABI are not the same and I didn't want to spend time putting together a common ABI file

src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
dohaki: Yea, you are right. Should be very unlikely

src/modules/web3/services/abi/WETH_1.json
dohaki: Hmm okay, got you.. Maybe we could also just reduce the ABI to the minimal interfaces with need / care about, i.e.
```
  {
    "anonymous": false,
    "inputs": [
      {
        "indexed": true,
        "name": "dst",
        "type": "address"
      },
      {
        "indexed": false,
        "name": "wad",
        "type": "uint256"
      }
    ],
    "name": "Deposit",
    "type": "event"
  },
```

PR #424, opened 2024-06-13
feat: arbitrum rebates program endpoints

melisaguevara:
This PR is adding the endpoints to query arbitrum rebates program data. It's also implementing cache for the following endpoint responses:
- rewards/earned
- rewards/arb-rebates/summary
- rewards/op-rebates/summary


src/modules/rewards/services/arb-rebate-service.ts
```diff
@@ -36,12 +41,110 @@ export class ArbRebateService {
...
+
+  public async getArbRebateRewards(query: GetRewardsQuery) {
+    const limit = parseInt(query.limit ?? "10");
+    const offset = parseInt(query.offset ?? "0");
+    const userAddress = assertValidAddress(query.userAddress);
+
+    const baseQuery = this.buildBaseQuery(this.arbRewardRepository.createQueryBuilder("r"), userAddress);
+
+    const rewardsQuery = baseQuery
+      .leftJoinAndSelect("r.rewardToken", "rewardToken")
```
amateima, line 124: Is this left join needed?

src/modules/rewards/services/arb-rebate-service.ts
```diff
@@ -36,12 +41,110 @@ export class ArbRebateService {
...
 
+  public async getEarnedRewards(userAddress: string) {
+    userAddress = assertValidAddress(userAddress);
+
+    const data = await this.cacheManager.get(getArbEarnedRewardsCacheKey(userAddress));
+    if (data) return data;
+
+    const baseQuery = this.buildBaseQuery(this.arbRewardRepository.createQueryBuilder("r"), userAddress);
+    const { arbRewards } = await baseQuery
+      .select("SUM(CAST(r.amount as DECIMAL))", "arbRewards")
```
amateima, line 59: I think here we should add a WHERE condition to filter only unclaimed rewards

src/modules/rewards/services/arb-rebate-service.ts
```diff
@@ -36,12 +41,110 @@ export class ArbRebateService {
...
     private ethProvidersService: EthProvidersService,
     private appConfig: AppConfig,
     private dataSource: DataSource,
   ) {}
 
+  public async getEarnedRewards(userAddress: string) {
+    userAddress = assertValidAddress(userAddress);
+
+    const data = await this.cacheManager.get(getArbEarnedRewardsCacheKey(userAddress));
+    if (data) return data;
```
amateima, line 55: Caching the data here would not bring many improvements because the earned referral rewards are not cached, so on every request you have to wait for them to be computed. What do you think if we cache the entire response of the `/rewards/earned` endpoint, not the sub-components

src/modules/rewards/adapter/db/arb-reward-fixture.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function mockArbRewardEntity(overrides: Partial<ArbReward>) {
+  return {
+    depositPrimaryKey: getRandomInt(),
+    recipient: "0x",
+    metadata: { rate: 0.95 } as RewardMetadata,
+    amount: "1000000000000000000",
+    amountUsd: "1",
+    rewardTokenId: 1,
+    rewardTokenPriceId: 1,
```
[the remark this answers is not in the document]
melisaguevara: No, you're right. I've deleted it

src/modules/rewards/services/arb-rebate-service.ts
melisaguevara: No! Good catch, I've removed it from this file and also from op-rebate-service

src/modules/rewards/services/arb-rebate-service.ts
melisaguevara: Yes, I agree, I thought we were using both unclaimed and claimed in the earned referral rewards service but we are not. I've updated both arb and op query.

src/modules/rewards/services/arb-rebate-service.ts
melisaguevara: Definitely, yes! I've implemented it and also moved the cache logict to the rewards service. Lmk what u think

amateima (approved):

PR #431, opened 2024-06-28
feat: including fill validation

melisaguevara:
We are now checking that the fill event hash matches de deposit event hash to discard invalid fills.

### DIFF
--- a/src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
@@ -285,12 +285,21 @@ export class BlocksEventsConsumer {
       } else if (event.args.relayExecutionInfo) {
         const typedEvent = event as FilledV3RelayEvent;
         const message: FillEventsV3QueueMessage = {
+          depositId: typedEvent.args.depositId,
+          depositor: typedEvent.args.depositor,
+          exclusiveRelayer: typedEvent.args.exclusiveRelayer,
+          exclusivityDeadline: typedEvent.args.exclusivityDeadline,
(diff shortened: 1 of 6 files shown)


src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
```diff
@@ -285,12 +285,21 @@ export class BlocksEventsConsumer {
...
+          exclusiveRelayer: typedEvent.args.exclusiveRelayer,
+          exclusivityDeadline: typedEvent.args.exclusivityDeadline,
+          fillDeadline: typedEvent.args.fillDeadline,
+          fillType: typedEvent.args.relayExecutionInfo.fillType,
+          originChainId: typedEvent.args.originChainId.toNumber(),
+          inputAmount: typedEvent.args.inputAmount,
+          inputToken: typedEvent.args.inputToken,
+          destinationChainId: chainId,
+          outputAmount: typedEvent.args.outputAmount,
+          outputToken: typedEvent.args.outputToken,
```
amateima, line 299: If old messages are present in the queue, some of these fields will be missing, so we need to discard the message

amateima (changes requested):

src/modules/scraper/adapter/messaging/BlocksEventsConsumer.ts
melisaguevara: oh yes, sorry, you mentioned this before.
When you say old messages you refer before V3?

amateima: no, I mean existing messages on the queue before this deployment

amateima (approved):

PR #444, opened 2024-07-08
feat: add blast indexing

dohaki:
Closes ACX-2087

### DIFF
--- a/src/modules/configuration/index.ts
+++ b/src/modules/configuration/index.ts
@@ -333,7 +333,7 @@ export const configValues = () => ({
           startBlockNumber: 5574280,
           abi: JSON.stringify(SpokePoolV3Abi),
           acrossVersion: AcrossContractsVersion.V3,
-          firstDepositId: 1,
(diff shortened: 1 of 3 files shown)


src/modules/configuration/index.ts
```diff
@@ -327,6 +327,24 @@ export const configValues = () => ({
           firstDepositId: 1,
         },
       ],
+      [ChainIds.blast]: [
+        {
+          address: "0x2D509190Ed0172ba588407D4c2df918F955Cc6E1",
+          startBlockNumber: 5574280,
+          abi: JSON.stringify(SpokePoolV3Abi),
+          acrossVersion: AcrossContractsVersion.V3,
+          firstDepositId: 1,
```
bmzig, line 336: The first deposit ID for these pools is 1,000,000

dohaki: fixed!

bmzig (approved):

PR #455, opened 2024-07-16
fix: fix blast bridge fee

amateima:


src/modules/scraper/service.ts
```diff
@@ -348,7 +349,14 @@ export class ScraperService {
   }
 
   public async fixBridgeFee() {
-    const deposits = await this.depositRepository.createQueryBuilder("d").where("d.bridgeFeePct < 0").getMany();
+    const deposits = await this.depositRepository.query(`
+      select d.id
+      from deposit d
+      inner join token it on d."tokenId" = it.id
+      inner join token ot on d."outputTokenId" = ot.id
+      where (d."sourceChainId" = 81457 and it.symbol = 'USDB' and ot.symbol = 'DAI')
```
james-a-morris, line 357: Unsure if this is common but would it make sense to parameterize this to not hardcode as 81457?

james-a-morris (approved):
Makes sense to me. I'm not super familiar with this repo but if all tests pass then LGTM. 

If possible/desired, maybe have another engineer look at this to confirm

src/modules/scraper/service.ts
amateima: This is only a temporary addition to fix the data for the existing deposits. Will be removed

PR #458, opened 2024-07-17
fix: process arb rewards claim events

amateima:
This PR adds support for indexing and processing claim events from arb rewards contract

### DIFF
--- a/src/modules/scraper/adapter/messaging/MerkleDistributorBlocksEventsConsumerV2.ts
+++ b/src/modules/scraper/adapter/messaging/MerkleDistributorBlocksEventsConsumerV2.ts
@@ -2,7 +2,7 @@ import { OnQueueFailed, Process, Processor } from "@nestjs/bull";
 import { Logger } from "@nestjs/common";
 import { Job } from "bull";
 import { InjectRepository } from "@nestjs/typeorm";
-import { Repository, QueryFailedError } from "typeorm";
+import { Repository, DataSource } from "typeorm";
 import { utils } from "ethers";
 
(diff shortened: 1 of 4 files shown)


src/modules/scraper/adapter/messaging/MerkleDistributorBlocksEventsConsumerV2.ts
```diff
@@ -25,38 +25,82 @@ export class MerkleDistributorBlocksEventsConsumerV2 {
...
-          ScraperQueue.MerkleDistributorClaim,
-          {
-            claimId: claim.id,
-          },
-        );
-      } catch (error) {
-        if (error instanceof QueryFailedError && error.driverError?.code === "23505") {
-          // Ignore duplicate key value violates unique constraint error.
-          this.logger.warn(error);
-        } else {
```
melisaguevara, line 53: we don't need to handle this anymore?

melisaguevara (approved):
LGTM!

src/modules/scraper/adapter/messaging/MerkleDistributorBlocksEventsConsumerV2.ts
amateima: no, because this queryBuilder has this nice feature [orIgnore()](https://github.com/across-protocol/scraper-api/pull/458/files#diff-2dd0f7e424c3af2b19a2f54b3d0823d8027b4ec481fc0b060aa49cef4ce33af7R92). If the insert query breaks a table constraint, it won't do the insert

PR #459, opened 2024-07-17
feat: add indexing of scroll

dohaki:
Closes ACX-2331


amateima (approved):
What tokens are we going to support on Scroll?

PR #465, opened 2024-07-19
feat: capping op and arb rebates to a maximum of 25bps

melisaguevara:


eslint.config.mjs
```diff
@@ -51,6 +51,7 @@ export default [
         },
       ],
       "@stylistic/comma-dangle": ["error", "always-multiline"],
+      "@stylistic/semi": ["error", "always"],
```
james-a-morris, line 54: Good call

src/modules/rewards/services/arb-rebate-service.ts
```diff
@@ -268,18 +267,25 @@ export class ArbRebateService {
...
       DateTime.fromJSDate(deposit.depositDate).minus({ days: 1 }).toJSDate(),
       rewardToken.symbol.toLowerCase(),
     );
+
+    const inputAmountUsd = new BigNumber(deposit.amount)
+      .multipliedBy(inputTokenPrice)
+      .dividedBy(new BigNumber(10).pow(deposit.token.decimals));
+    const maxRewardsUsd = inputAmountUsd.multipliedBy(REWARDS_PERCENTAGE_LIMIT);
     const rewardsUsd = new BigNumber(bridgeFeeUsd).multipliedBy(ARB_REBATE_RATE).toFixed();
+    const cappedRewardsUsd = BigNumber.min(maxRewardsUsd, rewardsUsd).toFixed();
```
james-a-morris, line 281: Does `.toFixed()` in this case default to some decimal precision?

src/modules/rewards/services/arb-rebate-service.ts
```diff
@@ -268,18 +267,25 @@ export class ArbRebateService {
...
 
+    const inputTokenPrice = deposit.price.usd;
     const historicRewardTokenPrice = await this.marketPriceService.getCachedHistoricMarketPrice(
       DateTime.fromJSDate(deposit.depositDate).minus({ days: 1 }).toJSDate(),
       rewardToken.symbol.toLowerCase(),
     );
+
+    const inputAmountUsd = new BigNumber(deposit.amount)
+      .multipliedBy(inputTokenPrice)
+      .dividedBy(new BigNumber(10).pow(deposit.token.decimals));
```
james-a-morris, line 278: Nit: Is this something we do a lot in the code? We may want to consider putting this into a function

src/modules/rewards/services/arb-rebate-service.ts
```diff
@@ -268,18 +267,25 @@ export class ArbRebateService {
...
+    const cappedRewardsUsd = BigNumber.min(maxRewardsUsd, rewardsUsd).toFixed();
     const rewardsAmount = ethers.utils.parseEther(
-      new BigNumber(rewardsUsd).dividedBy(historicRewardTokenPrice.usd).toFixed(18),
+      new BigNumber(cappedRewardsUsd).dividedBy(historicRewardTokenPrice.usd).toFixed(18),
     );
 
     return {
       rewardToken,
-      rewardsUsd,
+      rewardsUsd: cappedRewardsUsd,
```
james-a-morris, line 288: This looks to cap the reward amount in USD. Should we also want to cap the reward amount in terms of the token (for `rewardsAmount` below)

src/modules/scraper/adapter/messaging/MerkleDistributorBlocksEventsConsumerV2.ts
```diff
@@ -35,7 +35,7 @@ export class MerkleDistributorBlocksEventsConsumerV2 {
     const claimEvents = await this.getClaimEvents(filteredDistributorContracts, from, to);
     this.logNumberOfEventsFound(claimEvents, from, to);
     const insertResults = await this.insertClaimEvents(chainId, claimEvents);
-    const messages = insertResults.map((claim) => ({claimId: claim.id}))    
+    const messages = insertResults.map((claim) => ({claimId: claim.id}));
```
dohaki, line 38: hmm how did these get through CI? 🤔

dohaki (approved):
Looks good 👍

src/modules/scraper/adapter/messaging/MerkleDistributorBlocksEventsConsumerV2.ts
melisaguevara: It is related to [this config](https://github.com/across-protocol/scraper-api/pull/465/files/16f383a006d6fe15bd210b4950d1db9b7a8152fc#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R54) we just added missing

src/modules/rewards/services/arb-rebate-service.ts
melisaguevara: No, it doesn't

src/modules/rewards/services/arb-rebate-service.ts
melisaguevara: Yes, there are some places where we do this. I think I can work on this on a separate PR to change it in all the required places

src/modules/rewards/services/arb-rebate-service.ts
james-a-morris: What is the precision of using `toFixed` without any params?

src/modules/rewards/services/arb-rebate-service.ts
melisaguevara: Since `rewardsAmount` is calculated using `rewardsUsd` which is capped, I'd say there's no need to cap it in terms of the reward token. But lmk if I'm missing something here.

src/modules/rewards/services/arb-rebate-service.ts
melisaguevara: It returns the string representation of the number without rounding it. [This](https://mikemcl.github.io/bignumber.js/#toFix) are the docs for this method

james-a-morris (approved):

PR #466, opened 2024-07-22
fix: input token price for Blast USDB transactions rewards

melisaguevara:
At the contract level, the exchange ratio for USDB and DAI is 1:1. But as USDB and DAI have different prices, we price those transactions using always DAI to avoid inconsistencies  .

### DIFF
--- a/src/modules/deposit/model/deposit.entity.ts
+++ b/src/modules/deposit/model/deposit.entity.ts
@@ -76,7 +76,8 @@ export type FeeBreakdown = {
 
 export type PartialDeposit = Pick<
   Deposit,
-  "id" | "status" | "depositDate" | "amount" | "destinationChainId" | "sourceChainId" | "feeBreakdown" | "depositTxHash" | "price" | "token"
+  "id" | "amount" | "sourceChainId" | "destinationChainId" | "status" | "depositTxHash" | "depositDate" |
+  "feeBreakdown" | "token" | "price" | "outputToken" | "outputTokenPrice"
 >;
 
 @Entity()
--- a/src/modules/rewards/services/arb-rebate-service.ts
+++ b/src/modules/rewards/services/arb-rebate-service.ts
@@ -32,6 +32,8 @@ const partialDepositKeys: (keyof Deposit)[] = [
   "feeBreakdown",
 ];
 
+const requiredDepositRelations: (keyof Deposit)[] = ["token", "price", "outputToken", "outputTokenPrice"];
+
 @Injectable()
 export class ArbRebateService {
   private logger = new Logger(ArbRebateService.name);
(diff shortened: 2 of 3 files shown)


src/modules/deposit/model/deposit.entity.ts
```diff
@@ -76,7 +76,8 @@ export type FeeBreakdown = {
 
 export type PartialDeposit = Pick<
   Deposit,
-  "id" | "status" | "depositDate" | "amount" | "destinationChainId" | "sourceChainId" | "feeBreakdown" | "depositTxHash" | "price" | "token"
+  "id" | "amount" | "sourceChainId" | "destinationChainId" | "status" | "depositTxHash" | "depositDate" |
+  "feeBreakdown" | "token" | "price" | "outputToken" | "outputTokenPrice"
```
amateima, line 80: Not something actionable now, but I don't know if this is the place to put the `PartialDeposit` model as the name is too generic and its scope it's related tightly to the rewards service. 
Partial deposits are needed in different places and each use case has its own list of required fields, so I don't think we can have a single PartialDeposit fitting all the needs.

amateima (approved):

src/modules/rewards/services/arb-rebate-service.ts
```diff
@@ -320,6 +322,17 @@ export class ArbRebateService {
     }
   }
 
+  private getInputTokenPrice(deposit: PartialDeposit): string {
+    if (
+      deposit.sourceChainId === ChainIds.blast &&
+      deposit.token.symbol === "USDB" &&
+      deposit.outputToken.symbol === "DAI"
```
james-a-morris, line 329: Will this condition ever not be true with blast? Could we consider removing the check for `deposit.outputToken.symbol === "DAI"`?

src/modules/rewards/services/op-rebate-service.ts
```diff
@@ -337,6 +337,17 @@ export class OpRebateService {
...
+    if (
+      deposit.sourceChainId === ChainIds.blast &&
+      deposit.token.symbol === "USDB" &&
+      deposit.outputToken.symbol === "DAI"
+    ) {
+      return deposit.outputTokenPrice.usd;
+    }
+    return deposit.price.usd;
+  }
+
```
james-a-morris, line 350: This is a clone - can we not create a parent service with this logic & have both the OP/Arb classes inherit from it?

james-a-morris: cc @amateima as this could be an interesting refactoring opportunity for the next tech debt cycle 😄

melisaguevara: I think it would make sense to do a refactor when we are going to add another rebate service in the future.

src/modules/rewards/services/arb-rebate-service.ts
melisaguevara: My understanding is that at the contract level it is possible to set different input and output tokens so I think we should keep this check.

PR #469, opened 2024-07-25
fix: get prices only for FE supported tokens

amateima:


src/modules/market-price/services/service.ts
```diff
@@ -12,6 +12,13 @@ export class MarketPriceService {
     private coinGeckoService: CoinGeckoService,
   ) {}
 
+  /**
+   * Check if the token is supported by the CoinGecko API
+   */
+  public doesTokenSupportPricingApi(symbol: string): boolean {
+    return !!symbolIdMap[symbol.toLowerCase()];
+  }
+
```
melisaguevara, line 21: nit: could 'isTokenSupportedByPricingApi' be a better name for this func?

amateima: Yep 👍 Changed it!

PR #477, opened 2024-08-12
fix: missed fill event cron

melisaguevara:
The cron we use to look for missed fill events checks up to 1000 deposits each time. Since there are more than 1000 pending deposits in the database, for the thousand we were processing each run, we were founding only a dozen of fills. 
To avoid that, this PR modifies the query we do to order deposits by deposit date to check first the most recent ones.

It also fix a bug that kept suspended 'FindMissedFillEvent" jobs looping. Now they will stop running once the status is changed to suspended.

I'm still looking why we are missing so much fill events when we process blocks, specially on Base. But in the meantime, this should help us find more missed fills.

### DIFF
--- a/src/modules/scraper/adapter/messaging/FindMissedFillEventConsumer.ts
+++ b/src/modules/scraper/adapter/messaging/FindMissedFillEventConsumer.ts
@@ -78,6 +78,7 @@ export class FindMissedFillEventConsumer {
           { id: jobId },
           { status: FindMissedFillEventJobStatus.Suspended },
         );
+        return;
       }
     }
 
--- a/src/modules/scraper/service/CheckMissedFillEventsCron.ts
+++ b/src/modules/scraper/service/CheckMissedFillEventsCron.ts
@@ -54,7 +54,6 @@ export class CheckMissedFillEventsCron {
       .where("d.status = :status", { status: "pending" })
       .andWhere("d.outputTokenAddress is not null")
       .andWhere("d.depositDate <= NOW() - INTERVAL '5 minutes'")
-      .andWhere("d.fillDeadline >= NOW()") // ignore expired deposits
       .andWhere("j.id is null")
       .orderBy("d.depositDate", "DESC")
       .limit(1000)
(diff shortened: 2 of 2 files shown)


src/modules/scraper/service/CheckMissedFillEventsCron.ts
```diff
@@ -54,7 +54,9 @@ export class CheckMissedFillEventsCron {
       .where("d.status = :status", { status: "pending" })
       .andWhere("d.outputTokenAddress is not null")
       .andWhere("d.depositDate <= NOW() - INTERVAL '5 minutes'")
+      .andWhere("d.fillDeadline >= NOW()") // ignore expired deposits
       .andWhere("j.id is null")
+      .orderBy("d.depositDate", "DESC")
```
james-a-morris, line 59: ah ok to latest first. & we're dropping the expired deposits

src/modules/scraper/adapter/messaging/FindMissedFillEventConsumer.ts
```diff
@@ -78,6 +78,7 @@ export class FindMissedFillEventConsumer {
           { id: jobId },
           { status: FindMissedFillEventJobStatus.Suspended },
         );
+        return;
```
james-a-morris, line 81: is this necessary

james-a-morris (commented):
two comments

src/modules/scraper/adapter/messaging/FindMissedFillEventConsumer.ts
james-a-morris: Ah I see that we need it to fix that bug

melisaguevara: Yep. That's because we retry fail messages every minute. But for suspended jobs we don't want to retry.
So this suspended jobs are being retried over and over again.

src/modules/scraper/service/CheckMissedFillEventsCron.ts
melisaguevara: Yes exactly, we don't have a 'expired' status in the scraper so that's why i'm using fillDeadline.
But did you have any question regarding this?

james-a-morris (approved):
I think this looks good but we *may* want to hold off until we have a more comprehensive fix.

I will leave this to an engineer who is more familiar with the repo

melisaguevara:
> I think this looks good but we _may_ want to hold off until we have a more comprehensive fix.
> 
> I will leave this to an engineer who is more familiar with the repo

Thanks, James!
Dong-Ha is also OOO this week, so I'm merging only the ordering clause for now. I'll discuss this with Alex when he's back.

PR #478, opened 2024-08-19
fix: zora spoke config

melisaguevara:


src/modules/market-price/adapters/coingecko/index.ts
```diff
@@ -23,6 +23,7 @@ export const symbolIdMap = {
   arb: "arbitrum",
   usdb: "usdb",
   lsk: "lisk",
+  usdzc: "usd-coin-ethereum-bridged",
```
melisaguevara, line 26: There's no tracker for 'USDzC' in Coingecko, but I think we can use the same as for 'USDC.e'

amateima: Is this the reason for the increased number of retried messages in the `TokenDetails` queue? I was about to investigate that today

amateima (approved):

src/modules/market-price/adapters/coingecko/index.ts
melisaguevara: Should be solved after merging this.

PR #490, opened 2024-09-18
fix: use output address from hub pool events

amateima:


src/modules/deposit/service.ts
```diff
@@ -34,11 +36,58 @@ export class DepositService {
...
+    quoteTimestamp: number) {
+    const quoteDatetime = new Date(quoteTimestamp * 1000);
+    const mainnetBlockNumber = originChainId === ChainIds.mainnet
+      ? depositBlockNumber : (await this.blockRepository.findOne({
+        where: {
+          chainId: ChainIds.mainnet,
+          date: LessThanOrEqual(quoteDatetime),
+        },
+        order: { date: "DESC" },
+      })).blockNumber;
```
melisaguevara, line 61: @amateima There's an edge case with this approach. This will fail if you don't have a block matching the criteria on mainnet. Some workarounds I can think of are:
- Using the SDK's `BlockFinder` class. It has a `getBlockForTimestamp` function we can use instead of the Block table.
- Using DefiLlama's API. It has an endpoint that also resolves timestamps to block numbers. See `https://coins.llama.fi/block/{chain}/{timestamp}` docs [here](https://defillama.com/docs/api).
- Not a complete fix but we could also filter by blockNumber only when there's more than one rebalance route for a token.

I think this current logic will work in production as the block table has enough data (and will always do(?)) but I just wanted to flag this in case you consider this is something we should fix before merging.

melisaguevara:
Before merging, we have to set `ENABLE_HUB_POOL_EVENTS_PROCESSING` to true and wait until all the `SetPoolRebalanceRoute` events are stored.

PR #491, opened 2024-10-08
feat: add world chain support

dohaki:
Closes ACX-2716


src/modules/configuration/index.ts
```diff
@@ -396,6 +397,15 @@ export const configValues = () => ({
           firstDepositId: 1,
         },
       ],
+      [ChainIds.zora]: [
```
bmzig, line 400: Should be worldChain

dohaki: yea right

PR #496, opened 2024-10-23
feat: add lisk to op rewards

melisaguevara:


james-a-morris (approved):
Will we need to coordinate a launch time with the FE?

amateima (approved):

PR #499, opened 2024-11-01
feat: add aleph zero

melisaguevara:


src/modules/configuration/index.ts
```diff
@@ -406,6 +407,15 @@ export const configValues = () => ({
           firstDepositId: 1,
         },
       ],
+      [ChainIds.alephZero]: [
+        {
+          address: "0x09aea4b2242abC8bb4BB78D537A67a245A7bEC64",
+          startBlockNumber: 4112376,
+          abi: JSON.stringify(SpokePoolV3Abi),
+          acrossVersion: AcrossContractsVersion.V3,
+          firstDepositId: 0,
```
james-a-morris, line 416: Does this fit with the deploy we made?

amateima: Is `firstDepositId` 0 or 1?

amateima (commented):
Do we need to do anything about the step no3 from [here](https://github.com/across-protocol/scraper-api/tree/master?tab=readme-ov-file#add-new-spokepool-contracts)?

src/modules/configuration/index.ts
melisaguevara: > Does this fit with the deploy we made?
Yes, do you see any differences? @james-a-morris

melisaguevara: > Is firstDepositId 0 or 1?

We set it to 0, [here](https://github.com/across-protocol/contracts/pull/704/files#diff-96d466a4a26299aa39f73fd3c54562c8630ac95ed678f171a0cb619e8717bb91R9-R15). The first item of `initArgs` is the `initialDepositId`

melisaguevara:
> Do we need to do anything about the step no3 from [here](https://github.com/across-protocol/scraper-api/tree/master?tab=readme-ov-file#add-new-spokepool-contracts)?

Uh I missed that. I've set the same value that we use for arbitrum here [9ca3de7](https://github.com/across-protocol/scraper-api/pull/499/commits/9ca3de7d6588526430764f0e6c6cec6f4db1fb0a)

james-a-morris (commented):
lgtm pending @amateima's signoff

amateima (approved):

PR #511, opened 2025-01-12
feat: add cher support

amateima:


src/modules/configuration/index.ts
```diff
@@ -426,6 +427,15 @@ export const configValues = () => ({
           firstDepositId: 1,
         },
       ],
+      [ChainIds.cher]: [
+        {
+          address: "0x3baD7AD0728f9917d1Bf08af5782dCbD516cDd96",
+          startBlockNumber: 1709997,
+          abi: JSON.stringify(SpokePoolV3Abi),
+          acrossVersion: AcrossContractsVersion.V3,
+          firstDepositId: 0,
```
melisaguevara, line 436:
```suggestion
          firstDepositId: 1,
```

Based on https://github.com/across-protocol/contracts/pull/856/files#diff-b855c32b12c2e5c16a5e7965bee7264bb0ce10b2d030a64303323e4d8572dc89R10

james-a-morris: We should probably have made this be 0 in the deployment but this needs to be set to 1

src/modules/scraper/service.ts
```diff
@@ -253,6 +253,10 @@ export class ScraperService {
       return 15;
     }
 
+    if (chainId === ChainIds.cher) {
+      return 15;
+    }
+
```
james-a-morris, line 259: There looks to be a pattern here that OP chains follow this 15 block lookback. We should probably optimize this

